PR #3, opened 2015-01-09
Set bs pointer to NULL before deleting

michelsnippe:
On windows, whenever OBS was closed, a cef-bootstrap.exe process and the obs32.exe process kept running.
Setting the bs pointer to NULL before delete seemed to fix this issue.


michelsnippe:
I've closed it, because it seems to crash like this.

PR #4, opened 2015-01-11
Added CefShutdown() to make sure all Cef browsers get shutdown.

michelsnippe:
Solves an issue where obs would keep running because not all browsers were shutdown.


jp9000:
Sorry, the plugin didn't get updated for a long time.  I only just started looking at it, was the first time compiling it.

PR #27, opened 2016-08-16
Visibility JS Bindings

DDRBoxman:


notr1ch:
This seems to be based off an older commit, could you rebase it to HEAD so your changes are more clear?

DDRBoxman:
Sorry about that, accidentally included a few extra non-essential tweaks I made.

DDRBoxman:
Do we keep a version number for the browser plugin? I'd like to expose that in the js too.

OsirisNL:
No, there is no version number being exposed.

PR #29, opened 2016-08-17
Fix building under OSX after unicode file handling broke it

DDRBoxman:
http://utf8everywhere.org/

Adds [nowide](http://cppcms.com/files/nowide/html/) for easier unicode handling.


DDRBoxman:
See #31 for an alternative using boost

PR #31, opened 2016-09-02
Use boost locale when loading paths

DDRBoxman:
http://utf8everywhere.org/

This is an alternative to PR #29 that uses Boost Locale instead of the standalone nowide library. It's a bit cleaner honestly.


DDRBoxman:
Resolved in #29

PR #60, opened 2017-01-18
Revert C++ 11 changes to things relying on Semaphores

Xaymar:
std::condition_variable doesn't do what it's name actually suggests it does. It's more like a condition if anything, as it's not a variable that can be used as a condition at all.

Side note, with this the browser source refuses to load webpages in 32-bit if a debugger is attached. Unsure what the actual cause is, but normally this shouldn't happen.

### DIFF
--- a/obs-browser/browser-manager-base.cpp
+++ b/obs-browser/browser-manager-base.cpp
@@ -129,11 +129,11 @@ int BrowserManager::Impl::CreateBrowser(
 		const std::shared_ptr<BrowserListener> &browserListener)
 {
 	int browserIdentifier = 0;
-	std::mutex createMutex;
-	std::condition_variable createEvent;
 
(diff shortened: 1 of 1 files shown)


jp9000:
Unfortunately going to revert the original multi-threaded change commit due to the fact that multi-threaded message queue doesn't work on mac according to this CEF issue report from the maintainers:
https://bitbucket.org/chromiumembedded/cef/issues/745/multi_threaded_message_loop-yes-on-mac-os

Your change did fix the hang, but it would break OSX browser builds.

I managed to fix the original hang with the original code instead just via adding a startup event, see commit notes here in this commit:  https://github.com/kc5nra/obs-browser/commit/47084fc9e4fa10b5798289d076dc0a55e7ea3de4

Tested with various testers who were able to replicate the hang.

PR #81, opened 2017-06-16
Add support for CEF version 3071

Xaymar:
A new flag was added which specifies if defined protocols bypass Content-Security-Policy checks. This change will default that to false in order to protect users from hijacking (through ads or similar).

### DIFF
--- a/obs-browser/browser-manager-base.cpp
+++ b/obs-browser/browser-manager-base.cpp
@@ -1,4 +1,5 @@
 #include <util/platform.h>
+#include <include/cef_version.h>
 
 #include "browser-manager-base.hpp"
 #include "browser-task.hpp"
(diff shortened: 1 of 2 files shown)


obs-browser/browser-manager-base.cpp
```diff
@@ -144,7 +144,6 @@ int BrowserManager::Impl::CreateBrowser(
 				new BrowserClient(renderHandler, loadHandler, browserOBSBridge));
 
 		CefWindowInfo windowInfo;
-		windowInfo.transparent_painting_enabled = true;
```
DDRBoxman: We probably still need to keep this, or does it default to true?

Xaymar: It no longer exists as the default behavior is to render transparent now.

Xaymar: See: https://bitbucket.org/chromiumembedded/cef/src/1ebbf3c4da4c55ccbd91d384bbe8ff457bd0441e/include/internal/cef_win.h?at=3071&fileviewer=file-view-default#cef_win.h-83

Xaymar:
Updated with #if's for older version support. Building with 2987, 3029, 3071 and newer should work now.

PR #82, opened 2017-06-16
Work around a potential lockup with newer CEF versions

Xaymar:
For an unknown reason, os_event_wait will wait forever even though the event itself is set to be manual and should never wait. This works around this issue and makes it work properly.


Xaymar:
After talking with Jim, I've investigated one of the potential causes for this: the event is destroyed while we are waiting on it. Unfortunately in all test cases, os_event_wait immediately returns instead of waiting forever, so this issue is caused by something else - or its a really rare race condition with event destruction and waiting on it happening just at the right time.

PR #106, opened 2018-03-13
obs-browser: web config UI

zavitax:
This PR requires https://github.com/jp9000/obs-studio/pull/1221 as a pre-requisite which resolves several OBS stability issues.

- add menu item under tools -> Configure Streaming Services
- clicking on the new item opens a modal dialog with CEF
- the modal dialog exposes JS API to query OBS for available video codecs
- the API exposes a method to set up Scenes, Sources, Output and Streaming using JS configuration object
- the API currently supports creating Game, Browser and Video Capture sources
- add User-Agent product marker "Open Broadcast Studio (OBS)"
- add option to suspend video elements in inactive browser sources: this saves a tremendous amount of CPU cycles for invisible browser sources without having to shut down the browser
- add support for standard browser visibility API in inactive browser sources
- resolvea race condition with BrowserSource::Impl::Listener::OnDraw being called after the browser source has been disposed 

Currently available on Windows only.
Conditional compilation statements have been added to enforce this.


OsirisNL:
Could you perhaps group all those commits in more logically seperated PR's?

RytoEX:
I agree with @OsirisNL.  As it is, this PR is a bit large to digest and review.  Splitting bug-fixes from individual feature implementations and submitting separate PRs for them would be extremely helpful.  When you do that, look out for commits that revert your own changes, and see if you can get rid of commits that cancel each other out.

dodgepong:
This WCUI idea looks interesting but sort of seems wholly inappropriate to shoehorn into `obs-browser`. The Browser source right now is specifically for rendering HTML/CSS/JS into the OBS preview, not for rendering OBS UI.

We have discussed the idea of having "browser panels" for a while now, which would allow users to essentially dock a web page onto OBS to show things like stream chat or an online dashboard of sorts, but the idea of giving websites that are loaded inside the browser source access to modify OBS settings seems very concerning from a security standpoint, referring to the changes to the `obsstudio` window object in particular here.

The hard-coding for a specific "streaming service" seems undesirable here if this is to be used as a generalized application. Also, I disagree with the term "streaming service" in this context, as it can be easily confused with companies that provide actual streaming service like Twitch or YouTube. Also, the included form is entirely unlocalizable.

Basically, it looks like you started with a cool idea and took it way too far. If StreamElements wants a plugin that has the power to change sources and configure video settings, they should create their own separate plugin (maybe a fork of this one) so that they can avoid security concerns with allowing arbitrary websites to fiddle with OBS settings. Such a plugin could probably include a nicer featureset for their purposes as well, and might not even need to use a CEF-based UI. Perhaps it could even open the door to rendering nice overlays without the need for CEF entirely.

The parts of this I could see merged are the parts relating to rendering a browser as part of the OBS UI, but I personally would rather see that in context of allowing a user to specify a URL and then attach that browser to OBS as a docked panel, like the other panels in OBS.

Also, this does include some nice bugfixes, too -- another reason to split up the commits.

RytoEX:
> We have discussed the idea of having "browser panels" for a while now, which would allow users to essentially dock a web page onto OBS to show things like stream chat or an online dashboard of sorts

Not to derail, but I actually made a branch last year that tried to embed CEF into a dockable element.  It kinda worked, but there were issues I couldn't solve at the time, owing to the complexities of CEF and rendering.  I eventually set it aside in favor of other projects.  In all likelihood, it's totally possible, but would need the hand of someone who is more familiar with those aspects.


Now, back to this PR...  I definitely agree with @dodgepong's security concerns.  Arbitrary user-specified websites probably shouldn't have access to OBS settings or user settings (like stream keys).  It seems too easy to abuse.

zavitax:
Let me try and address your concerns.

> This WCUI idea looks interesting but sort of seems wholly inappropriate to shoehorn into obs-browser. The Browser source right now is specifically for rendering HTML/CSS/JS into the OBS preview, not for rendering OBS UI.

This is very counter-intuitive, and yet, when considering the alternatives, obs-browser seems the right place to put this functionality in. Consider this: adding another plug-in with CEF would open the door for version conflicts, increased memory footprint and increased installation footprint. Adding a plug-in with another embedded browser would increase the memory & installation footprint temporarily off-setting version conflicts.

> We have discussed the idea of having "browser panels" for a while now, which would allow users to essentially dock a web page onto OBS to show things like stream chat or an online dashboard of sorts, but the idea of giving websites that are loaded inside the browser source access to modify OBS settings seems very concerning from a security standpoint, referring to the changes to the obsstudio window object in particular here.

> The hard-coding for a specific "streaming service" seems undesirable here if this is to be used as a generalized application.

> Now, back to this PR... I definitely agree with @dodgepong's security concerns. Arbitrary user-specified websites probably shouldn't have access to OBS settings or user settings (like stream keys). It seems too easy to abuse.

> The parts of this I could see merged are the parts relating to rendering a browser as part of the OBS UI, but I personally would rather see that in context of allowing a user to specify a URL and then attach that browser to OBS as a docked panel, like the other panels in OBS.

From security standpoint, the extended window.obsstudio JS object functionality is available within the WCUI pop-up dialog only. This is intentional: it allows displaying a list of services pre-baked into the bundled HTML file (which should provide a good balance between hard-coded and easily modifiable) which also displays a warning that clicking on the links will take the user to an external website. Hopefully, this is an acceptable balance between functionality (allowing the user to easily configure streaming environment with a few mouse clicks) and security (minimize the attack surface, keeping the list of available services under the control of the community, write-only access to settings).

The idea of "browser panels" or "browser widgets" is next on my list to tackle. The fixes available in this PR pave the way for doing so.

> Also, I disagree with the term "streaming service" in this context, as it can be easily confused with companies that provide actual streaming service like Twitch or YouTube.

Perhaps better wording would be appropriate. Suggestions?

> Also, the included form is entirely unlocalizable.

The included form is composed entirely from a list of services and a short warning message. The rest is external. Perhaps the warning message could be localized. Which mechanism would be preferred to localize it?

> Basically, it looks like you started with a cool idea and took it way too far. If StreamElements wants a plugin that has the power to change sources and configure video settings, they should create their own separate plugin (maybe a fork of this one) so that they can avoid security concerns with allowing arbitrary websites to fiddle with OBS settings. Such a plugin could probably include a nicer featureset for their purposes as well, and might not even need to use a CEF-based UI. Perhaps it could even open the door to rendering nice overlays without the need for CEF entirely.

The idea is to have the whole community benefit from the added functionality and not limit it to StreamElements or any other provider.

> Could you perhaps group all those commits in more logically seperated PR's?

I'm available to answer any questions

WizardCM:
I agree with all three people above - this PR should be split up - at the very least, bug fixes and features should be separate so they can be reviewed independently. It's possible for features to depend on bug fixes - but not the other way round.

The other benefit of splitting would be that certain things which are smaller, like bug fixes, would be reviewed and merged more quickly, which could pave the way for other developers to also submit their own PRs for other features that benefit from said bug fixes, without having to wait for the other pile of features that are much larger and therefore will take much longer to review.

dodgepong:
So, I've spent quite a while thinking about this PR and the functionality goals that it's trying to accomplish, and I think I've figured out what would be a more ideal way to implement it.

Ultimately, the problem that this PR is attempting to solve is the problem of how to allow third parties to change OBS settings and scenes/sources. For services like StreamElements, StreamLabs, etc. it can be a complicated process for users to add a browser source with the right size and URL to the right scenes, and a feature like this would smooth that process out.

When looking at the general problem of third party settings management, I actually find the solution in this PR to be too constricting. Any service that wants to be able to modify OBS settings would have to be added manually to this system by OBS maintainers, which means we're now maintaining a list of people "allowed" to use this feature, which sounds like a bad idea to me.

Instead, a more generalized solution is one that has existed in idea form for quite some time: a custom URI handler. Essentially, a site like StreamElements could have a button that users click on that is in the form of something like `obs://<url stuff here>` which OBS is then set to open and interpret (asking the user's permission, of course). So a user could go onto the overlay service's main site, change the things they need to, then click a button to add it all to OBS, sort of like what this has here.

Then there's the browser panel feature proposal, which would allow a user to essentially attach a browser window to the OBS UI in order to have things like stream chat or overlay dashboards inside the program. A user could add an overlay's site settings panel to OBS, then the user could modify their settings as needed, then press a button to add those things to OBS.

However, when being run from inside a browser panel, it seems a little awkward to have to interact with OBS via a URI handler still. You're inside OBS already, so why not talk to OBS directly?

This is where I think the idea of an expanded JavaScript API actually makes more sense. With such a JavaScript API, sites loaded into OBS via a browser panel could interact with OBS and do the things they need to do directly, provided that the user has granted permission to that panel to modify OBS settings. This permission could be granted either upon panel creation (i.e. a checkbox saying "Allow this site to change my OBS settings") or ask for permission via pop-up when the site attempts to change settings ("This site wants to change your OBS settings, Allow? Y/N" + a checkbox to remember this decision).

The last piece to this puzzle, then, is how to make the process of adding this panel to OBS seamless to users, because otherwise they are trading the problem of creating a browser source for the problem of creating a browser panel. The solution here, again, would be the URI handler. With a custom URI, a site could have a button that says something like "Open this site in OBS", and clicking that button would send a new URI request that OBS can handle to automatically create the browser panel pointing to the right URL. This eliminates the need for OBS to maintain a list of sites that support this feature, but still keeps it easy to discover and set up for users of those services.

In summary, I think a better implementation of this feature requires 2 things: 1) a custom URI handler that lets sites send settings directly to OBS via a `obs://` protocol (or whatever it will be named), and 2) an implementation of browser panels with an extended JavaScript API that the user can permit to change settings. Then, the workflow for a user of an overlay service would be:

* Go to overlay service site
* Click "Open this site in OBS" button, which is a custom protocol that OBS is set up to handle requests to, which would add a browser panel to OBS containing the appropriate site
* User sets things up the way they want for their overlays
* User clicks a button to apply settings, which fires JS events to set the settings in OBS
* Browser panel intercepts these requests and asks the user if they grant permission for these settings to be set, and lets them through if permission is granted.

zavitax:
> When looking at the general problem of third party settings management, I actually find the solution in this PR to be too constricting. Any service that wants to be able to modify OBS settings would have to be added manually to this system by OBS maintainers, which means we're now maintaining a list of people "allowed" to use this feature, which sounds like a bad idea to me.

> Instead, a more generalized solution is one that has existed in idea form for quite some time: a custom URI handler.

I've spent some time thinking about those valid points, while also considering UX and the added value of having a list of easily-accessible services available.
I think I found an approach which would both allow good UX and keep the solution completely open: we could maintain a dynamic list of available providers, while allowing to easily extend it with a URL handler scheme.

This would shoot several birds at once:

- Deliver good, clear, streamlined UX for setting up OBS environment by external services
- Be easily extensible via URL scheme handler mechanism, not limiting the services to a pre-baked list of "allowed" people
- Would enable services to add browser widgets as part of the delivered configuration in future versions
- Still maintain a good level of security by not allowing configuration changes by pages loaded from totally arbitrary URLs

So the proposed modification is:

- Move services list into a dynamic configuration file, keeping the popup dialog
- Add services to the list via URL scheme handler
- Add services list management to the configuration dialog

dodgepong:
When you say "we could maintain a dynamic list of available providers", are you referring to a list of providers hosted on an OBS server, or that the user would maintain their own list of providers locally to be extended?

zavitax:
> When you say "we could maintain a dynamic list of available providers", are you referring to a list of providers hosted on an OBS server, or that the user would maintain their own list of providers locally to be extended?

I'm referring to a local list maintained by the user

dodgepong:
At the very least, browser panels could (should) maintain a history of all the different URLs that it has been asked to open. I'm not convinced that OBS should have the concept of "overlay service providers" coded into the program, though.

To be clear, my proposal above involves scrapping the WCUI additions in this PR in favor of merging some of that functionality into browser panels.

PR #108, opened 2018-03-17
obs-browser: bugfix: don't paint invisible sources

zavitax:
WasHidden(false) prevents OnDraw() calls from being
made on hidden sources.
This saves CPU cycles on sources which are hidden.


jp9000:
Code has already been applied.

PR #113, opened 2018-05-03
obs-browser: do not paint for released BrowserSources

zavitax:
BrowserClient::OnPaint() and BrowserClient::OnAcceleratedPaint() events
can arrive even after calling BrowserSource::DestroyBrowser().
This adds a check whether the BrowserClient identified by the
CefBrowser passed to OnPaint() and OnAcceleratedPaint() is still
valid without introducing global locks.

### DIFF
--- a/browser-client.hpp
+++ b/browser-client.hpp
@@ -42,8 +42,13 @@ public:
 	CefRect popupRect;
 	CefRect originalPopupRect;
 
+	// This flag is checked by BrowserClient::OnPaint and
+	// BrowserClient::OnAcceleratedPaint
+	// If its value is false, the paint events are discarded.
+	bool isValid;
+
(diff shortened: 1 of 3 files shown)


browser-client.hpp
```diff
@@ -42,8 +42,13 @@ class BrowserClient : public CefClient,
 	CefRect popupRect;
 	CefRect originalPopupRect;
 
+	// This flag is checked by BrowserClient::OnPaint and
+	// BrowserClient::OnAcceleratedPaint
+	// If it's value is false, the paint events are discarded.
```
RytoEX: Spelling:  "it's" should be "its" (nitpick)

zavitax:
Spelling fixed @RytoEX

PR #119, opened 2018-07-23
RFC: Add install directives, allow dynamic runtime, and modular frontend-api

computerquip-streamlabs:
Currently, this is within a single commit but originally, I had them split up into two commits. This PR is mostly to seek interest. This commit does three things:

1) It gets rid of the copy commands and replaces them with cmake install logic. This allows the use of cpack and install targets to work correctly. I'm almost tempted to consider this a bug fix given I can't use the source without it. 

2) It adds back in the MT/MD switch which is required to compile against libcef when we use the dll_wrapper built with /MD. I'd perhaps consider this a bug as well. I'd guess currently compiled obs-browser doesn't compile against the dynamically-linked runtime version of libcef_dll_wrapper which may not be ideal. I'd imagine linker warnings are given when compiling concerning a mismatch.

3) It adds a new compile-time flag to control the frontend API logic. This logic is specific to the Qt frontend and doesn't work for anyone not shipping with the frontend API. 

Let me know and I'll reformat with wanted changes.


computerquip-streamlabs:
I also forgot that I revamped the FindCEF logic as well to be a little bit more friendly and readable in my opinion. Not really the goal here but could also be included if wanted.

PR #120, opened 2018-07-23
Log plugin version on module load

RytoEX:
In the browser source rewrite, the log line showing the plugin version number seems to have gotten lost.  I personally think all plugins should display their version number clearly in the log, so I'd like to see this added back in.  I've changed the log message a bit to be more similar to other plugins' log messages.

I'm not sure if it would be better to include browser-version.h in browser-app.hpp instead of directly including it here (and in browser-app.cpp, apparently).

Compiled and tested a 64-bit build on Windows 10 64-bit.


RytoEX:
Thanks for merging this!

PR #124, opened 2018-08-10
Revise FindCEF to use imported targets

computerquip-streamlabs:
This uses imported targets instead of storing things into variables. There's a few things that are great about imported targets:
1. You don't have to worry about Debug/Release when linking against libraries. You can link against the same target and get the correct version of library. 

2. You don't need to worry about include directories. If you link against the target, Cmake knows you're going to need the include directories to be part of the target you're building in order to use the library.

3. They're easier to specify and safer to use. If the target isn't correctly defined, CMake will give an error where as an empty variable is perfectly acceptable otherwise.

One thing this won't help with is installation. Currently, I'm using install/cpack style rules in order to build libobs distributions. The current CMake script doesn't work for this as it uses copy commands into rundir rather than install commands (so they'll just be excluded from a distribution or cpack build).

This script now works with single-config build systems (make or ninja for instance). The old script didn't because it assumes the Debug/Release paths generated by the default configuration types. It still does to some extent but now the library linked will depend on what's specified. For instance, if it finds the library in "Debug", and our current build type is "Debug", it will use the Debug library. Otherwise, if it only found the Release library, it will use the Release library. If you specify a folder made with Ninja, it will always be the _RELEASE variant and consequently the one always linked against (since there's no practical difference anyways).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,17 +23,14 @@ configure_file(
 	"${CMAKE_CURRENT_BINARY_DIR}/browser-config.h")
 
 include_directories("${CMAKE_CURRENT_BINARY_DIR}")
-include_directories("${CEF_ROOT_DIR}")
 
 # ----------------------------------------------------------------------------
 
 set(obs-browser_LIBRARIES
+	CEF::CEF
 	libobs
 	obs-frontend-api
 	)
-
-list(APPEND obs-browser_LIBRARIES
-	${CEF_LIBRARIES})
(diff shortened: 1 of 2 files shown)


computerquip-streamlabs:
I'll add a note that this probably shouldn't be merged *as is*. I've not tested it very well and it could break use cases that I'm not aware of.

computerquip-streamlabs:
Closing for now due to conflicts.

PR #125, opened 2018-08-10
Add option to use static runtime

computerquip-streamlabs:
There appears to have been misconception of not being able to use /MD when using libcef. This isn't true due to the use of the libcef_dll_wrapper described [here](https://bitbucket.org/chromiumembedded/cef/wiki/LinkingDifferentRunTimeLibraries).

This adds an option to dynamically link against the runtime by setting BROWSER_USE_STATIC_CRT to a falsey value. 

While I've set the default value here to true, I don't suggest statically against the runtime here at all. I've set it to default only because this is how it was already set and it felt less intrusive.


PatTheMav:
@WizardCM I think this can be closed - the underlying CMake implementation has obviously changed a lot in the last 7(!) years, but we also use the static CEF wrapper by default anyway (which is also independent of how CEF itself is linked/loaded anyway, as correctly pointed out).

PR #127, opened 2018-08-13
Define options as on/off values rather than enabled values

computerquip-streamlabs:
37095e5  contains a detailed description of why this is useful. In particular though, if you specify YES for example which is a valid true value in Cmake, it will fail because there isn't a corresponding macro. This just throws all of that out the window, let's Cmake check the value for true or false itself, and only then defines a variable to be set in the config.h file. If it's defined, it will be set to 1 (which I use since it's used with #if and not #if defined() statements) otherwise will be set to 0.

This replaces #121 in favor of a better changeset.


computerquip-streamlabs:
There's a bug here concerning the if check there so I'm closing this for now. Next time we update our browser source, I'll likely restage and make a new PR.

PR #128, opened 2018-08-13
Use install commands for consistency

computerquip-streamlabs:
This adds install commands which eases shipping shipping libobs as a distribution or cpack.


computerquip-streamlabs:
I don't use this in our code base right now and it's not maintained. Closing.

PR #143, opened 2019-01-07
Use fake UI for media stream

### LINKED ISSUE #146: Cannot use user media devices without flag
No way to get access to user media devices (webcams and microphones) without having to use flag. Added a flag in PR #143. Would love to see this in the next release.

hunterpollpeter:
Use fake UI for media stream avoids the need to grant camera/microphone permissions.

Fixes #146

### DIFF
--- a/browser-app.cpp
+++ b/browser-app.cpp
@@ -53,6 +53,8 @@ void BrowserApp::OnBeforeCommandLineProcessing(
 	}
 
 	command_line->AppendSwitch("enable-system-flash");
+	
+	command_line->AppendSwitch("use-fake-ui-for-media-stream");
 
 	command_line->AppendSwitchWithValue("autoplay-policy",
 			"no-user-gesture-required");


dodgepong:
While I don't necessarily disagree with this feature being enabled, can you describe your use case for this?

mphacker:
Just realized that the command lines from OBS are passed in to the browser source.  So although this would be a good addition I see it being more safe to pass in the parameter.   Having it built in could allow a rogue page to listen in to a user's input device with out them being aware.

WizardCM:
I agree this should be opt-in rather than the norm. As OBS passes flags to the browser source, this particular change shouldn't be merged.

hunterpollpeter:
My use case is creating overlays that use the webcam from the browser source itself. The idea is to make it easier for less technically inclined people to have an overlay with populated data.

hunterpollpeter:
I do see how this can be exploited and maybe hardcoding the flag isn't the best option. Would it be a better solution to make this an option on the browser-source configuration window? Or maybe even a preference within the application? Thanks for the discussion!

WizardCM:
I agree this would be best as an option. "Allow this page to access audio/video devices, and my display"

mnill:
I need this feature to make face masks plugin with browser, not with separate dll which is hard to install.

eliasthompson:
Been having issues with arguments like `--enable-media-stream` and `--use-fake-ui-for-media-stream` not working for browser sources when passed in v26. Has there been any traction on this PR in the past year or so? And if the option lies in each browser source properties, how would we be able to enable this for browser docks?

RytoEX:
> Been having issues with arguments like `--enable-media-stream` and `--use-fake-ui-for-media-stream` not working for browser sources when passed in v26.

@eliasthompson 
Can you provide a reproducible test case?  What specific platform/OS and what specific version of OBS?  What exactly are you trying to do, what is happening, and what do you expect to happen?  Though, this conversation should probably start in one of our support channels and be filed as its own issue if the flags are somehow not working as intended.  I'd rather this PR discussion stay focused on the merits of whether this PR should be merged or not.


> Has there been any traction on this PR in the past year or so? 

No, there has been no movement on this PR.  Per @WizardCM 's [comment above](https://github.com/obsproject/obs-browser/pull/143#issuecomment-458362077):
> I agree this should be opt-in rather than the norm. As OBS passes flags to the browser source, this particular change shouldn't be merged.

And his [other comment](https://github.com/obsproject/obs-browser/pull/143#issuecomment-569465234):
> I agree this would be best as an option in each source. "Allow this page to access audio/video devices, and my display"

We pass all user-specified flags to CEF.  I agree that we should not enable this flag by default for all users.


> And if the option lies in each browser source properties, how would we be able to enable this for browser docks?

You wouldn't, as far as I know.  Switches passed to CEF affect the process/application from startup to close.  I don't think they can be selectively applied to different tabs/windows (browser sources or browser panels, in our case).


Between the fact that we don't want to universally allow browser sources to automatically capture a user's audio or video without permission and the fact that a user can already pass any flag to CEF via an application shortcut, I'm in favor of closing this.

eliasthompson:
Yeah I didn't mean to derail the discussion with my own issues, I'll post those in the appropriate channels if they persist.

In terms of the PR, yeah, if the only way it works is by global enabling it for the during that the program is run, then not enabling it by default is the right call imo and the PR should be closed. But if there is any way that it can be set/manipulated while the program is running, I'd be highly in favor of some sort of GUI solution.

RytoEX:
> In terms of the PR, yeah, if the only way it works is by global enabling it for the during that the program is run, then not enabling it by default is the right call imo and the PR should be closed. But if there is any way that it can be set/manipulated while the program is running, I'd be highly in favor of some sort of GUI solution.

As far as I know, all CEF switches/flags are global.  There are some, like hardware acceleration, that we can check the value of and then do other things like create UI elements.  This, however, is a flag that bypasses a user permission dialog and prevents it from being displayed and automatically uses default audio and video devices for media streams.  I don't know of a way we could intercept the functionality and change it on a per source basis.

RytoEX:
Per conversation with @WizardCM about unaddressed concerns, I'm going to close this PR.

PR #150, opened 2019-02-17
Add replay buffer events to API

WizardCM:
Goes well with the `getStatus` function which returns the replay buffer status.


WizardCM:
Rebased with clang-format.

PR #153, opened 2019-02-19
Add JS events for source active/visible change

### LINKED ISSUE #72: Inconsistancy with JS hooks
# 1 - Passed Data Typings
With the `obsSceneChanged` event, `<event>.detail` is a JS object, similar to:  
```js
{height:1080,name:"brb",width:1920}
```

But with `window.obsstudio.getCurrentScene()` when the callback is called, the `data` argument is a non-parsed JSON **string**, resulting in `data` having to be passed through `JSON.parse()` before it is usable:
```js
// data as String: {"height":1080,"name":"brb","width":1920}
window.obsstudio.getCurrentScene((data) => {
    console.log(data); // results in a string
    console.log(JSON.parse(data)); // results in a js object
});
```

I feel the JSON data of the latter should be parsed into a JS object as it is with the former.  

# 2 - Different design patterns for similar functionality  
The interface is all over the place on design for similar functionality sets. For example `.onVisibilityChange` is a property that gets set with a handler function while other events use the `window.addEventListener()` methodology.

# 3 - Incompleteness  
Scenes have a `getCurrentScene` method and an `onSceneChange` event, but `onVisibilityChange` doesn't have an accompanying getter function; nor do the `obsStreaming*` and `obsRecording*` events.

Win10 - Home Edition  
OBS Studio: 18.0.1
BrowserSource: v1.29.0

WizardCM:
Partially fixes #72

Known issues:
- [x] These new events fire on all browsers, even those unrelated to the visibility change

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ window.addEventListener('obsSceneChanged', function(evt) {
 });
 ```
 #### Other events that are available
+* obsSourceVisibleChanged
+* obsSourceActiveChanged
 * obsStreamingStarting
 * obsStreamingStarted
 * obsStreamingStopping
--- a/obs-browser-plugin.cpp
+++ b/obs-browser-plugin.cpp
@@ -388,7 +388,8 @@ void RegisterBrowserSource()
 
 /* ========================================================================= */
 
-extern void DispatchJSEvent(std::string eventName, std::string jsonString);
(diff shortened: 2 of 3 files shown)


WizardCM:
Thanks to significant help from @Andersama, I got this to work much better.

WizardCM:
I've updated the PR to be compatible with recent changes, and ran a clang-format.

PR #155, opened 2019-03-06
obs-browser: flush panel cookie store on exit

zavitax:
If cookie store is not flushed, state might become inconsistent:
reading from the cookies file will become impossible.
No errors will be returned, yet no information will be available,
and no information will be stored: the cookie manager will appear to be
empty at all times and the `Cookies` file modified date will not update
moving forward.

This will result, for example, in Twitch chat window re-opening for
users which connected their Twitch account, but not having the
`auth-token` cookie available: once the user will try to send a
message to the chat, they will be redirected to Twitch login screen.


PatTheMav:
@zavitax Is this still an issue with the current version of OBS Studio?

PR #156, opened 2019-03-06
Set default size to size of canvas

cg2121:
I haven't tested this since I'm on Linux, but it should work.


WizardCM:
Closing this even though it's older (and untested), as #178 is more descriptive.

PR #157, opened 2019-03-14
Add audio capture for browser source

pkviet:
Relies on two cef commits (both on branch 3683):
58e1149 : Add ability to capture audio output to buffer via cef handler
1501768 : Mute audio in the browser (issue #1806)

This PR turns a browser source into a regular audio source, with a vue-meter on the Sound Mixer.

The audio from browser source is now muted to desktop.
It can be captured and monitored.
The PR is compatible with multiple audio sources in same browser page.

Main author: Osiris.
Other authors: Andersama & pkv

**In memoriam. Osiris**
Thanks for all that you did for this project. We deeply miss the
opportunity to keep working with you. - andersama , pkv & WizardCM

Co-authored-by: Michel Snippe <michel@sokar.nl>
Co-authored-by: Andersama <anderson.john.alexander@gmail.com>
Co-authored-by: pkv <pkv.stream@gmail.com>
Co-authored-by: WizardCM <git@wizardcm.com>

### DIFF
--- a/browser-client.cpp
+++ b/browser-client.cpp
@@ -22,6 +22,7 @@
 #include "json11/json11.hpp"
 #include <obs-frontend-api.h>
 #include <obs.hpp>
+#include <util/platform.h>
 
 using namespace json11;
 
@@ -60,6 +61,13 @@ CefRefPtr<CefContextMenuHandler> BrowserClient::GetContextMenuHandler()
 {
 	return this;
 }
+
+#if CHROME_VERSION_BUILD >= 3683
+CefRefPtr<CefAudioHandler> BrowserClient::GetAudioHandler()
+{
+	return this;
+}
+#endif
 
 bool BrowserClient::OnBeforePopup(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame>,
 				  const CefString &, const CefString &,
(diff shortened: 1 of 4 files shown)


WizardCM:
Note: this requires heavy testing with [CEF 3683 (Minimal Distribution)](http://opensource.spotify.com/cefbuilds/index.html) before merge, across all platforms.

pkviet:
**update**
- I've added checks for cef 3683.
- Note that the first commit is required for cef >= 3683 and has a separate PR #158. Once the other commit is merged, I'll rebase to get rid of the duplicate commit.
- added WizardCM as author, since he helped fix the compilation with cef 3683.

pkviet:
**update**
- i've tested on cef master where both audio commits are merged as well as hardware acceleration fixes.
Unfortunately, browser acceleration is still broken ...
- the only build which has both audio and hardware acceleration is a custom patched build by Osiris.
We'll have to wait until this is fixed in cef.

pkviet:
**update**
- this PR for cef fixes hardware acceleration for cef 3683 and later (3729) : https://bitbucket.org/chromiumembedded/cef/pull-requests/222/fix-use_external_begin_frame-support-and/diff
- I've compiled cef master with this patch and obs-browser has both audio and hardware acceleration fully working.
- However, a recent commit in master (chromium 3729) breaks api with CookieManager. Therefore panels are not working any more.
I have a separate fix for that here: https://gist.github.com/pkviet/3932d59e159ca4c8d03c89d228ddabcc
It's not PR'd because it requires adding some ifdefs to check cef version.
It would be ackward though to ifdef so many blocks of code and make for less readibility. 
Maybe it'd be better to remove support for earlier versions of cef ...
That's something which will require some input from Jim.

RytoEX:
>       It would be ackward though to ifdef so many blocks of code and make for less readibility.
>       Maybe it'd be better to remove support for earlier versions of cef ...
>       That's something which will require some input from Jim.

Personally, I'm totally for dropping official support for much older versions of CEF.  I think the oldest version we should support (ifdefs, code paths, etc.) are the version we use for official obs-browser/obs-studio builds, or the oldest "Current Release Branch" of CEF, whichever is older.  If the commits to remove older version support are clear and restricted to just dropping support for one version at a time, then it should be possible for others to revert those so that they can use older CEF builds if they want.  As you noted, this is something Jim would have to decide.  If we do decide to drop support for older CEF versions, I'd be happy to go through the codebase and remove sections specific to those dropped older versions.

pkviet:
NB:
the CookieManager fix requires changes obs UI too , cf : https://gist.github.com/pkviet/4a0fac632c835352d4f158d14641147e

SuslikV:
> ...
> It would be ackward though to ifdef so many blocks of code and make for less readibility.
> ...

Such small files? Is in't better to if def them in include? And give a new name like "blah-blah-blah-3729.hpp" or similar?

Show-vars:
Any good news? I apologize for the useless question, but I think many people are interested to know the current situation.

pkviet:
> Any good news? I apologize for the useless question, but I think many people are interested to know the current situation.

yes and no.
The fix for hardware acceleration was merged into cef master BUT not to branches; there are issues with debug builds that the author has not addressed yet.
I tested release builds on branches with the fix and it seems to work (audio + hardware acceleration).
So we still need to wait. Another PR is in the works which rehauls the hardware acceleration.

pkviet:
**update June 19 2019**
Some very good news.
The hardware acceleration fix has now been committed by Marshall in branch 3770.
(It's not in any other branches.)
But branch 3770 introduces a series of API changes; I've added commits to fix them.
I checked that branch 3770 builds both with browser audio and hardware acceleration.
And that Twitch oauth is still working.

pkviet:
**update July 12 2019**
Applied clang-format to the PR.
One question though is that the PR is littered by  CHROME_VERSION_BUILD checks.
This makes the code unnecessarily convoluted.
It would be simpler as discussed above to decide to support only the last cef release.

WizardCM:
I agree that starting with these changes, we should support 3770 at the minimum going forward. There is no reason to support older builds. This would require testing macOS with the latest CEF too.

pkviet:
**update July 13**
- removed the commits related to cef 3770 fixes (Jim implemented them separately)
- removed the cef 3683 fix (merged by Jim)
- rebased against current master, works perfectly with cef 3770

dragonbane0:
I pray that this gets merged soon - we need it badly lol

pkviet:
**update**
- fixed a timestamp issue (was in microseconds instead of nanoseconds due to an API change when the cef PR was merged in master). Thanks to Jim for solving the bug.

PR #158, opened 2019-03-15
Fix AddCustomScheme function for cef >= 3683

pkviet:
Since cef 3683 AddCustomScheme has  only two args instead of 6 and uses a bitmask arg.
This fixes the function for cef >= 3683.


pkviet:
pr updated by using defs CEF_SCHEME_OPTION_CORS_ENABLED (= 1<<4) and CEF_SCHEME_OPTION_STANDARD = 1 << 0 instead of the raw numbers

PR #163, opened 2019-05-18
Panels: Right click Reload, use Qt context menu

WizardCM:
### Description

This PR introduces two core changes, split into their own two commits **(Gif examples at the end)**.

1. It allows for the modification of the contents of the CEF context menu. This is a native capability, and I've used it to perform two minor changes: the first is to enable a Reload option, to refresh the page. The second is to hide the Print option, as it's not needed. I considered also hiding "View page source". Feedback here would be good. This change overall is quite small, and can be used down the line to add/remove any options, whether they're native ones (like the Reload one) or completely custom functions.

2. Next I added a function to replace the native CEF context menu with a Qt menu. This is also a native capability, and ensures that not only the context menu is consistent with others, but also matches the OBS theme.

The second change went through a few iterations, but in the end I settled on this.

### Motivation and Context

The primary motivation is consistency. The context menu for browser docks doesn't currently follow the same design rules as all other context menus in OBS, making it feel out of place and "tacked on". It also behaves unexpectedly when the user presses the context menu key on the keyboard, where it opens two menus where one cannot be closed.
The second motivation is extensibility. By using native CEF functions to modify the dropdown, it opens the doors to further options in the future: like a "Dock Configuration" button, for example. In this case, it starts off simple by adding a much-requested Reload option.

Implementation details:

* `RunContextMenu` requires a "return true" if you're rendering your own context menu (which, we are). The callback can be run synchronously or asynchronously, and has no time limit
* CEF Docs state: "Do not keep references to `params` or `model` outside of this callback." therefore I had to create a new object to store the basics of the contents of the dropdown. The `model` gives us everything we need - the `text`, `command id`, `enabled` status, and menu item `type`.
* Qt doesn't allow you to create widgets outside of the UI thread, so once the menu items are acquired from the `model`, we create a `QMenu` in the UI thread, and populate it with the items, creating separators as needed.
* The use of `exec()` was chosen as it's used in other places in OBS, and because it allowed for the smallest bit of code: it synchronously returns whether an action was clicked or not, allowing the ability to then run the necessary CEF callback. **If you don't run the Cancel() callback if an item hasn't been chosen, then you can't open the dropdown a second time.** The event just never triggers.

### How Has This Been Tested?

Add a custom browser dock, or enable service integration, and right click anywhere within the webpage. Depending on whether text is selected, and what element you right click on, CEF will give you different options in the context menu. Example: right click a random area where there's no text, and then right click a text input field (the latter will unlock Copy/Paste).

**Before:**

**After:**

### Types of changes

### Checklist:

### DIFF
--- a/panel/browser-panel-client.cpp
+++ b/panel/browser-panel-client.cpp
@@ -3,6 +3,9 @@
 
 #include <QUrl>
 #include <QDesktopServices>
+#include <QApplication>
+#include <QMenu>
+#include <QThread>
 
 #include <obs-module.h>
 #ifdef _WIN32
@@ -204,6 +207,57 @@ void QCefBrowserClient::OnBeforeContextMenu(CefRefPtr<CefBrowser>,
 	if (model->IsVisible(MENU_ID_PRINT)) {
 		model->Remove(MENU_ID_PRINT);
 	}
+}
+
+bool QCefBrowserClient::RunContextMenu(
+	CefRefPtr<CefBrowser>, CefRefPtr<CefFrame>,
+	CefRefPtr<CefContextMenuParams>, CefRefPtr<CefMenuModel> model,
+	CefRefPtr<CefRunContextMenuCallback> callback)
+{
+	std::vector<std::tuple<std::string, int, bool, int>> menu_items;
+	menu_items.reserve(model->GetCount());
+	for (int i = 0; i < model->GetCount(); i++) {
+		menu_items.push_back(
+			{model->GetLabelAt(i), model->GetCommandIdAt(i),
+			 model->IsEnabledAt(i), model->GetTypeAt(i)});
+	}
+
+	QMetaObject::invokeMethod(
+		QCoreApplication::instance()->thread(),
+		[menu_items, callback]() {
+			QMenu contextMenu;
+			std::string name;
+			int command_id;
+			bool enabled;
+			int type_id;
+
+			for (int i = 0; i < menu_items.size(); i++) {
+				std::tie(name, command_id, enabled, type_id) =
+					menu_items[i];
+				switch (type_id) {
+				case MENUITEMTYPE_COMMAND: {
+					QAction *item =
+						new QAction(name.c_str());
+					item->setEnabled(enabled);
+					item->setProperty("cmd_id", command_id);
+					contextMenu.addAction(item);
+				} break;
+				case MENUITEMTYPE_SEPARATOR:
+					contextMenu.addSeparator();
+					break;
+				}
+			}
+
+			QAction *action = contextMenu.exec(QCursor::pos());
+			if (action) {
+				QVariant cmdId = action->property("cmd_id");
+				callback.get()->Continue(cmdId.toInt(),
+							 EVENTFLAG_NONE);
+			} else {
+				callback.get()->Cancel();
+			}
+		});
+	return true;
 }
 
 void QCefBrowserClient::OnLoadEnd(CefRefPtr<CefBrowser>,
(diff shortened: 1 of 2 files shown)


WizardCM:
Rebased with clang-format.

panel/browser-panel-client.cpp
```diff
@@ -186,6 +194,73 @@ bool QCefBrowserClient::OnBeforePopup(
 	return true;
 }
 
+void QCefBrowserClient::OnBeforeContextMenu(CefRefPtr<CefBrowser>,
+					    CefRefPtr<CefFrame>,
+					    CefRefPtr<CefContextMenuParams>,
+					    CefRefPtr<CefMenuModel> model)
+{
+	if (model->IsVisible(MENU_ID_BACK) &&
+	    model->IsVisible(MENU_ID_FORWARD) &&
```
RytoEX, line 203: You can probably only check one of these conditions, or check if "Copy" does not exist, since the intent was to not add the Reload button when text is highlighted.

panel/browser-panel-client.cpp
```diff
@@ -186,6 +194,73 @@ bool QCefBrowserClient::OnBeforePopup(
...
+	    (!model->IsVisible(MENU_ID_RELOAD) &&
+	     !model->IsVisible(MENU_ID_RELOAD_NOCACHE))) {
+		model->InsertItemAt(2, MENU_ID_RELOAD_NOCACHE, "&Reload");
+		if (model->IsVisible(MENU_ID_PRINT)) {
+			model->Remove(MENU_ID_PRINT);
+		}
```
RytoEX, line 209: I don't think you need to nest this here if the intent is to _always_ remove the Print menu item.

Also, do we actually need to check if Print is visible?  Does it throw an exception/error if you try to remove print if it's _not_ visible?

panel/browser-panel-client.cpp
```diff
@@ -186,6 +194,73 @@ bool QCefBrowserClient::OnBeforePopup(
...
+
+			QAction *action = contextMenu.exec(QCursor::pos());
+			if (action) {
+				QVariant cmdId = action->property("cmd_id");
+				callback.get()->Continue(cmdId.toInt(),
+							 EVENTFLAG_NONE);
+			} else {
+				callback.get()->Cancel();
+			}
+		});
```
RytoEX, line 260: Discussed with @WizardCM , but I'd split this into a separate function and call it similar to how we use `QMetaObject::invokeMethod` elsewhere.  That should help with indenting and line length.

RytoEX (changes requested):
Discussed on call with @WizardCM .  Leaving review notes here to help keep a paper trail.

panel/browser-panel-client.cpp
WizardCM: Done

panel/browser-panel-client.cpp
WizardCM: Done. Even if it doesn't currently throw an exception, I'd rather check before deleting.

panel/browser-panel-client.cpp
WizardCM: I attempted this, but `invokeMethod` from a non-Qt thread can only invoke slots/signals from QThread.

RytoEX: Is this then creating an anonymous slot/signal?

WizardCM: Pretty much?

jp9000:
What happens if a URL overwrites the right-click command?

WizardCM:
@jp9000 Solid question! I have verified [on this page](https://codepen.io/SitePoint/full/MYLoWY) that the page's custom context menu takes priority in the correct situations. By the looks of it the CEF code never gets hit, so V8 (Chrome's JavaScript engine) probably handles that internally.

Issue #164, opened 2019-06-01
Add an option to refresh cache when set to visible.


Etuldan:
Like the 
https://github.com/obsproject/obs-browser/blob/d6805f53d0c09694807a974e0371bd3ca04cb177/obs-browser-plugin.cpp#L399-L405 
, could we have similar feature for 
https://github.com/obsproject/obs-browser/blob/d6805f53d0c09694807a974e0371bd3ca04cb177/obs-browser-plugin.cpp#L391-L394 ?

I mean, force refresh cache when set to visible ?
Will be very helpful when manipulating from websocket. (see https://github.com/Palakis/obs-websocket/issues/339 )

Regards

PR #165, opened 2019-06-01
Option to refresh browser on set visible.

Etuldan:
Closes https://github.com/obsproject/obs-browser/issues/164

Add an option to enable refresh of browser when source set to visible.

Only locales for en-US and fr-FR


Vainock:
Please don't translate something here, do that in the OBS Studio Crowdin project: https://crowdin.com/project/obs-studio. Please remove the translation of the string.

Etuldan:
So, I have to close this one, revert on my repo (or start from scratch) the FR-fr translation (I'm native French, that's why), and create a new PR  ?

Vainock:
I understand why do you want to translate it, but it is completey unnecessary because your translations here will not be uploaded to Crowdin and have to be made again there.

I'm not pretty familiar with Git and this stuff, but I think you only have to revert the changes you made in the fr file and commit again, but ask a developer for sure.

kkartaltepe:
There is no need to close this PR. Simply fix up your commit and force push it.

Etuldan:
> 
> 
> There is no need to close this PR. Simply fix up your commit and force push it.

Thx, done :)
(and I fixed also indentation (from spaces to tabs)à

obs-browser-plugin.cpp
```diff
@@ -390,7 +392,10 @@ void RegisterBrowserSource()
 	};
 	info.show = [] (void *data)
 	{
-		static_cast<BrowserSource *>(data)->SetShowing(true);
+		BrowserSource *bs = static_cast<BrowserSource *>(data);
+		if (bs->estart_on_visible)
```
megalon: I haven't tested this, but this looks like a typo to me.
Shouldn't this be `restart_on_visible`, not `estart_on_visible` ?

Etuldan: Indeed ...
I'll fix this

obs-browser-plugin.cpp
```diff
@@ -390,7 +392,10 @@ void RegisterBrowserSource()
 	};
 	info.show = [] (void *data)
 	{
-		static_cast<BrowserSource *>(data)->SetShowing(true);
+		BrowserSource *bs = static_cast<BrowserSource *>(data);
+		if (bs->restart_on_visible)
+			bs->Refresh();
```
WizardCM: The option you're likely looking for is actually `bs->cefBrowser->ReloadIgnoreCache();` if you're going off the #164 request.

PR #168, opened 2019-09-04
obs-browser: simulate page visibility API

zavitax:
### Description, Motivation and Context

WasHidden() CEF API was abandoned on CEF 3507+ by commit b7d3ac2 due to
a bug with OSR in Chromium. This defeated the standard Page Visibility
API which webpages rely upon to conserve resources when browser sources
are invisible.

We simulate the Page Visibility API by updating the relevant document
properties (document.hidden & document.visibilityState) and firing
the 'visibilitychange' event when browser source visibility state
updates.

We also address initial visible/hidden state by storing the state
reported to SetShowing() and reporting initial state immediately after
creating the browser (initial SetShowing call is made before browser is
created due to the async nature of this process).

### How Has This Been Tested?
Page Visibility API was tested with local HTML/JS pages. Tested both changes to page visibility and initial state.

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -37,6 +37,27 @@ extern bool QueueCEFTask(std::function<void()> task);
 static mutex browser_list_mutex;
 static BrowserSource *first_browser = nullptr;
 
+static void SendBrowserVisibility(CefRefPtr<CefBrowser> browser, bool isVisible)
+{
+	if (!browser)
+		return;
+
+#if ENABLE_WASHIDDEN
+	if (isVisible) {
+		browser->GetHost()->WasHidden(false);
+		browser->GetHost()->Invalidate(PET_VIEW);
+	} else {
+		browser->GetHost()->WasHidden(true);
+	}
+#endif
+
+	CefRefPtr<CefProcessMessage> msg =
+		CefProcessMessage::Create("Visibility");
+	CefRefPtr<CefListValue> args = msg->GetArgumentList();
+	args->SetBool(0, isVisible);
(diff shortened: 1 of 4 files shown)


obs-browser-source.cpp
```diff
@@ -37,6 +37,27 @@ extern bool QueueCEFTask(std::function<void()> task);
...
+		return;
+
+#if ENABLE_WASHIDDEN
+	if (isVisible) {
+		browser->GetHost()->WasHidden(false);
+		browser->GetHost()->Invalidate(PET_VIEW);
+	} else {
+		browser->GetHost()->WasHidden(true);
+	}
+#endif
```
jp9000: Hm, shouldn't it just do an #else for the code below this?  If ENABLE_WASHIDDEN is true shouldn't the emulation be unnecessary?  (I admit I might be misunderstanding)

zavitax: @jp9000 the back end of this sits on top of the OBS Visibility API ("Visibility" message which in turn triggers "onVisibilityChange"), this moves creation and dispatching of the "Visibility" message from SetShowing() into this separate function, and if available, makes use of WasHidden() as well.

zavitax: as a bonus, this simplifies the code in SetShowing()

zavitax: if you move on to browser-app.cpp, you'll notice that actual emulation is indeed disabled when WAS_HIDDEN=1

jp9000:
Great work!

PR #170, opened 2019-09-07
Enable large local media file access

zavitax:
### Description, Motivation and Context
In newer CEF builds, the http://absolute/ scheme handler factory seems
to never complete when accessing large media files due to an apparent
bug in the Chromium media player (XHR requests complete fine).
At the same time, file:// URLs are working just fine.

file:// URLs will be used where available (CEF 3440+) falling back to
`BrowserSchemeHandlerFactory` on older CEF versions.

When file:// URLs are in effect, we'll disable `web_security` to allow
CORS requests to remote APIs, preserving functionality previously
enabled by a custom browser scheme handler.

`BrowserSchemeHandlerFactory` is simplified by using CEF built in
`CefStreamResourceHandler` in accord with `CefStreamReader` instead of
manually implementing a `CefResourceHandler`.

### How Has This Been Tested?
Tested with direct access to local media file (WEBM) and access via a large HTML page with a &lt;video /&gt; media player built against CEF 3440 and CEF 3770.

### Types of changes

### Checklist:


RytoEX:
I'm a little concerned about _completely_ disabling CEF's web security for browser sources loading local files instead of either enabling CORS on a custom scheme for local files or opening up access for file URLs.  Would it be acceptable to enable `file_access_from_file_urls` and `universal_access_from_file_urls` instead of disabling `web_security`, or is there some use case where those do not loosen restrictions enough?

According to the CEF docs and comments from Marshall, the best option is to use a custom scheme, but in lieu of that, I made [a branch that uses the `*_from_file_urls` settings](https://github.com/RytoEX/obs-browser/tree/access_from_file_urls).  With that branch, I tested loading large local media files, and I tested a basic CORS request, and the code seems to work fine for the CEF version from the OBS CI (cef_binary_75.1.16+g16a67c4+chromium-75.0.3770.100_windows64_minimal).  I didn't want to jump straight to a PR without discussion, but if it would be better to submit a PR to create a space for discussing and testing it, I'm okay with that too.

zavitax:
> I'm a little concerned about _completely_ disabling CEF's web security for browser sources loading local files instead of either enabling CORS on a custom scheme for local files or opening up access for file URLs. Would it be acceptable to enable `file_access_from_file_urls` and `universal_access_from_file_urls` instead of disabling `web_security`, or is there some use case where those do not loosen restrictions enough?

If we're certain we don't have a gap in test cases to make sure it's enough, I don't see a reason not to tighten security.

PR #177, opened 2019-10-01
obs-browser: Replace register with __restrict

jpark37:
### Description
Replace register with __restrict since register is not supported in C++17.

### Motivation and Context
Trying to enable C++17 as a byproduct of wanting to support WinRT headers.

### How Has This Been Tested?
Breakpoint inspection.

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-source-audio.cpp
+++ b/obs-browser-source-audio.cpp
@@ -25,15 +25,16 @@ void BrowserSource::EnumAudioStreams(obs_source_enum_proc_t cb, void *param)
 	}
 }
 
-static inline void mix_audio(float *p_out, float *p_in, size_t pos,
+static inline void mix_audio(float *__restrict p_out,
+			     const float *__restrict p_in, size_t pos,
 			     size_t count)
 {
-	register float *out = p_out;
-	register float *in = p_in + pos;
-	register float *end = in + count;
+	float *__restrict out = p_out;
+	const float *__restrict in = p_in + pos;
+	const float *__restrict end = in + count;
 
 	while (in < end)
-		*(out++) += *(in++);
+		*out++ += *in++;
 }
 
 bool BrowserSource::AudioMix(uint64_t *ts_out,


jpark37:
TIL parenthesis don't matter.

Xaymar:
Are C++ `register` (deprecated since C++11 IIRC) and C `__restrict` behaving the same way? I am aware that `register` only works in some compilers, and may just be ignored as the compiler usually makes the correct choices. To my knowledge, it (like inline) is just meant as a hint to the compiler to generate certain behavior.

jpark37:
They mean different things, so I didn't mean to imply that the two meant the same thing.

The register keyword was meant as a hint to use a register for a local rather than spill to stack. C++17 assumes compilers are so good at managing this that the keyword is removed entirely, and not just deprecated. This particular usage is curious; how can you have a pointer to a register? Registers aren't addressable.

__restrict (standardized as restrict in C but not C++) is meant to hint to a compiler that all restricted pointers in a function more or less never alias each other. In this situation, the compiler is free to unroll a bit and consider more than one float at a time, which it can't do if the arrays are allowed to overlap. (I didn't check if compilers will optimize this particular loop, but they could in the future if they don't now.) This function is a candidate for SIMD as well, but I didn't want to go that far.

jpark37:
Overlapping memory
----
	Memory = [3][4][5]
	In     = ^-----
	Out    =    ^-----

**Unoptimized mix**
Read In0: 3
Read/Add/Write Out0: 4+3=7
Read In1: 7
Read/Add/Write Out1: 5+7=12
Final = [3][7][12]

**Attempt 2 element unroll**
Read In0, In1: 3,4
Read/Add/Write Out0, Out1: 4,5 + 3,4 = 7,9
Final = [3][7][9]

Outputs are not the same. Compiler cannot unroll.

__restrict, promise no overlap
----
	In  = [3][4] ...
	Out =        ... [4][5]

**Unoptimized mix**
Read In0: 3
Read/Add/Write Out0: 4+3=7
Read In1: 4
Read/Add/Write Out1: 5+4=9
Final = [7][9]

**Attempt 2 element unroll**
Read In0, In1: 3,4
Read/Add/Write Out0, Out1: 4,5 + 3,4 = 7,9
Final = [7][9]

Outputs are the same. Compiler is able to unroll.

PR #178, opened 2019-10-10
Use OBS canvas size as browser source default size

RytoEX:
### Description
This is a minor QoL change to make browser sources default to be the same size as the user's OBS canvas. For users using overlays, this should reduce the work required to add a browser source for an overlay. For all other use cases, this should not create extra work.

If the canvas width and height are not defined (I don't know why this would happen), the width and height should fallback to 800x600. Feedback on this particular section of code would be appreciated.

**This PR depends on https://github.com/obsproject/obs-studio/pull/2073/ to maintain backwards compatibility with previously created browser sources that did not have a width and height different from the default specified.**

image

### Motivation and Context
At TwitchCon, some users expressed that having browser sources default to the same size as the OBS canvas would save them time/work for setting up scenes using browser source based overlays.

### How Has This Been Tested?
I compiled and tested this on Windows 10 Home 1903 (Build 18362.388) 64-bit.
1. Open OBS.
2. Add a new browser source.

Before this change, a new browser source defaults to width 800 and height 600.
After this change, a new browser source defaults to the width and height of the OBS canvas.

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-plugin.cpp
+++ b/obs-browser-plugin.cpp
@@ -110,10 +110,15 @@ overflow: hidden; \
 
 static void browser_source_get_defaults(obs_data_t *settings)
 {
+	struct obs_video_info ovi = {0};
+	obs_get_video_info(&ovi);
+	int width = ovi.base_width ? ovi.base_width : 800;
+	int height = ovi.base_height ? ovi.base_height : 600;
+
 	obs_data_set_default_string(settings, "url",
 				    "https://obsproject.com/browser-source");
-	obs_data_set_default_int(settings, "width", 800);
-	obs_data_set_default_int(settings, "height", 600);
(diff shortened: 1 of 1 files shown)


WizardCM:
Challenge: this migrates all browser sources where the user hasn't changed the size from their previous size of 800x600 to the new size based on their canvas.

WizardCM:
As the ticket that this PR is dependent on has been cancelled/put on hold, I'm considering closing this PR for now (and likely pinning an Issue for it instead). Thoughts @cg2121 @RytoEX?

WizardCM:
Closing this in favour of #202

12345678kkmmxx:
[](https://discord.com/channels/348973006581923840/387365167110815744/707695798585917480)

PR #188, opened 2019-12-28
Display errors in browser panels instead of an empty page

WizardCM:
### Description
When a browser panel fails to load, show an error screen instead of a blank page. This uses the load handler built into CEF.

Text for lines 1, 2 and the link at the bottom are **always** translated via the locale.
Line 3 uses a locale translation if it's available, otherwise it falls back to the raw error code.

In terms of the HTML/CSS itself, it's near-identical to https://obsproject.com/browser-source - minus the white border, and with a smaller logo. This should make it easy to keep the designs in sync. The logo is an SVG so that it can load offline.

### Motivation and Context

While it makes sense for a browser source to turn blank if a webpage fails to load, it does not make sense for a browser panel. With this PR, the user knows why the page failed to load, and has an easy link to reload the page.

### How Has This Been Tested?
* Create a browser panel pointed to an invalid url
* After the timeout period, the error page should be displayed

### Types of changes

### Checklist:

### DIFF
--- a/data/error.html
+++ b/data/error.html
@@ -1,0 +1,180 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+	<head>
+		<title>OBS | Browser Source</title>
+		
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width,initial-scale=1.0">
+		<meta name="description" content="OBS (Open Broadcaster Software) is free and open source software for video recording and live streaming. Stream to Twitch, YouTube and many other providers or record your own videos with high quality H264 / AAC encoding.">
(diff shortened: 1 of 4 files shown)


dodgepong:
I realize this is something that should have been discussed when the PR was made rather than after it was merged, but my gut reaction to this is that it should have an option to be disabled, and disabled by default. If something happens during a production then it's possible that the whole screen is suddenly filled with an OBS error message. I think it's a good idea to be able to show errors to the user but not like this during a production.

dodgepong:
My mistake, this is just for browser panels. Ignore me.

PR #189, opened 2019-12-29
Fix CSS injection by using a style element

WizardCM:
### Description

Uses a `<style>` element, and bas64 encoded text that gets converted back to a string at runtime, to ensure all characters are escaped properly when transferred from C++ to JavaScript.

### Motivation and Context

Custom CSS doesn't load properly on many pages due to their configuration (for valid reason).
`obs-browser: Refused to load the stylesheet 'data:text/css;charset=utf-8;base64,' because it violates the following Content Security Policy directive: "style-src 'self' 'unsafe-inline' ". Note that 'style-src-elem' was not explicitly set, so 'style-src' is used as a fallback.`

By using a `<style>` element instead of a `<link>`, the CSS is injected safely and consistently.

The reason it didn't load before this PR is because by loading in a `<link>` element, the CSS is considered external - even though it's being loaded 100% internally within the browser source, and without a network call.

### How Has This Been Tested?

Define some custom CSS, like the below, and load it on the default `obsproject.com/browser-source` page. Notice without this change it doesn't load, and with this change, it loads.
```css
html {
    border: solid 10px red;
}
body:before {
    content: "My quotes are properly escaped";
}
```

### Types of changes

### Checklist:

### DIFF
--- a/browser-client.cpp
+++ b/browser-client.cpp
@@ -365,19 +365,12 @@ void BrowserClient::OnLoadEnd(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame> frame,
 	if (frame->IsMain()) {
 		std::string base64EncodedCSS = base64_encode(bs->css);
 
-		std::string href;
-		href += "data:text/css;charset=utf-8;base64,";
-		href += base64EncodedCSS;
-
 		std::string script;
-		script += "var link = document.createElement('link');";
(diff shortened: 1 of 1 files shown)


Xaymar:
Do you have some URLs where this would work and could be tested on?

WizardCM:
The default `obsproject.com/browser-source` is a good place to start, in my experience. I think some alert providers may have a similar Content Security Policy.

PR #193, opened 2020-01-26
Unblacklist Intel igpu in single GPU configs

pkviet:
### Description
The blacklist of Intel igpu was motivated by issues with dual gpu setups in laptops.
A few people with old Intel igpu (which unfortunately have not  been documented) had also reported
issues with hardware acceleration in single gpu setups. Intel igpus were therefore blacklisted.
This however prevents the hardware decoding of webm videos by Intel igpu in single gpu setups.
With this commit hardware decoding of webm will be possible for Intel igpu in single gpu setups
while people who had issues with hardware acceleration can still disable it.

### Motivation and Context
Re-enables hardware acceleration for Intel igpu in single gpu setups.
They are still blacklisted in dual gpu setups.
Fixes [https://github.com/obsproject/obs-studio/issues/2248](https://github.com/obsproject/obs-studio/issues/2248).

### How Has This Been Tested?
Tested on a laptop with core i5 8250u and HD620 iGPU + win 10 Home.

### Types of changes
- Bug fix (hardware decoding of webm by Intel igpu)
- Breaking change (might perhaps cause issues for very old Intel igpus; these users can fix their issue by disabling hardware acceleration, in the rare odds they resurface.) 

### Checklist:

### DIFF
--- a/obs-browser-plugin.cpp
+++ b/obs-browser-plugin.cpp
@@ -500,6 +500,34 @@ static const wchar_t *blacklisted_devices[] = {
 	L"Intel", L"Microsoft", L"Radeon HD 8850M", L"Radeon HD 7660", nullptr};
 #endif
 
+static inline bool is_intel(const std::wstring &str)
+{
+	return wstrstri(str.c_str(), L"Intel") != 0;
+}
+
+static void check_hwaccel_support(void)
+{
+	/* do not use hardware acceleration if a blacklisted device is the
+	 * default and on 2 or more adapters */
+	const wchar_t **device = blacklisted_devices;
+
+	if (adapterCount >= 2 || !is_intel(deviceId)) {
+		while (*device) {
+			if (!!wstrstri(deviceId.c_str(), *device)) {
+				hwaccel = false;
+				blog(LOG_INFO, "[obs-browser]: "
+					       "Blacklisted device "
+					       "detected, "
+					       "disabling browser "
+					       "source hardware "
+					       "acceleration.");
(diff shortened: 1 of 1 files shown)


obs-browser-plugin.cpp
```diff
@@ -524,15 +524,19 @@ bool obs_module_load(void)
...
+				if (!!wstrstri(deviceId.c_str(), *device)) {
+					hwaccel = false;
+					blog(LOG_INFO, "[obs-browser]: "
+						       "Blacklisted device "
+						       "detected, "
+						       "disabling browser "
+						       "source hardware "
+						       "acceleration.");
+					break;
+				}
```
notr1ch: The condition should probably be moved outside of the loop as it isn't influenced by the loop variables.

pkviet: done, thanks R1ch

energizerfellow:
Might want to add a checkbox at the bottom of Advanced Settings, that's enabled by default, about disabling Intel GPU hardware video decoding acceleration if multiple GPUs are detected. It would be nice to at least have that as an option for the support peeps.  Put is under a section at the bottom of Advanced Settings called "Quirks" or something like that.

It would be nice to see a similar checkbox for the NVENC blacklist, but I digress.

notr1ch:
There is already an option to disable browser source hardware acceleration, adding a second one which only takes effect with multiple GPUs would just be confusing. I doubt there are many people frequently switching between single and multi GPU setups to necessitate this.

PR #202, opened 2020-02-22
obs-browser: Update default size

cg2121:
### Description
Changes default size to size of canvas.

### Motivation and Context
A better default.

### How Has This Been Tested?
Created browser source.

### Types of changes

### Checklist:


obs-browser-plugin.cpp
```diff
@@ -126,6 +126,27 @@ static void browser_source_get_defaults(obs_data_t *settings)
 	obs_data_set_default_bool(settings, "reroute_audio", false);
 }
 
+static void browser_source_get_defaults_v2(obs_data_t *settings)
```
WizardCM: Could this technically call `browser_source_get_defaults()` first, then set the width/height in this function?

RytoEX:
I prefer something like this, but it's probably down to personal preference.
https://github.com/RytoEX/obs-browser/tree/change-default-size

notr1ch:
Do we even need a v2 here? We're not changing anything except the default value so users shouldn't experience any unexpected issues as the defaults are only used on source creation or reset.

In the event a user does restore the browser source back to defaults (and is thus presumably OK with losing all their settings and getting new defaults), it would be confusing that the defaults of a legacy browser source do not match the defaults of a newly added browser source, with no visible way to distinguish between them.

WizardCM:
Yes, v2 is required because the defaults aren't saved on source creation. It's the reason v2 exists at all.

WizardCM (changes requested):
The feedback I provided in my previous review has not yet been responded to, so I don't believe this should be merged as-is.

notr1ch:
Ah I see, I didn't realise our defaults worked like that.

obs-browser-plugin.cpp
notr1ch: I was also thinking this as code duplication is generally bad, but after realising our defaults work in the way they do, it does also seem somewhat fitting to keep them separated. If we're committed to never changing v1 defaults then I think removing the duplication is OK.

obs-browser-plugin.cpp
```diff
@@ -126,6 +126,27 @@ static void browser_source_get_defaults(obs_data_t *settings)
...
+{
+	struct obs_video_info ovi;
+	obs_get_video_info(&ovi);
+
+	obs_data_set_default_string(settings, "url",
+				    "https://obsproject.com/browser-source");
+	obs_data_set_default_int(settings, "width", ovi.base_width);
+	obs_data_set_default_int(settings, "height", ovi.base_height);
+	obs_data_set_default_int(settings, "fps", 30);
```
notr1ch, line 138: Would it also make sense to default to the canvas FPS here?

obs-browser-plugin.cpp
cg2121: Updated function to remove duplicate code.

obs-browser-plugin.cpp
cg2121: This fps setting is only used if 'use custom fps' is checked.

notr1ch: I'm aware of that, but in those situations wouldn't it also make sense to default to the current FPS?

WizardCM:
So, extending what I said above, today I learned that if you create a browser source and click "OK", *all* settings are saved because of the way the functions in the browser source are written (this should not actually be the case, but it is). So it's safe to assume that any browser made since the v2 rewrite (at least) has the current default resolution saved.

tt2468:
Fair enough. I do believe in the meta tags idea, so I'll try to find time to work on that.

Issue #200, opened 2020-02-16
Interact window provides incorrect codes on keypress and keydown events


wooferzfg:
Whenever a `keydown` or `keypress` event gets triggered, `event.code` gets set to the wrong value.

## Expected Behavior

Add the following scripts and then hold down the A key when having any window selected. It should log `KeyA` for each of the events.

![chrome_Fmef3ORunR](https://user-images.githubusercontent.com/8262173/74608312-97831e80-50a5-11ea-9246-f7e1a61eaa53.png)

![chrome_QmuZEl5Ii5](https://user-images.githubusercontent.com/8262173/74608313-97831e80-50a5-11ea-9dcc-ef5a0c3a2720.png)

## Current Behavior

Add the following scripts and then hold down the A key when having the Browser Source Interact window selected. It logs `F7` and an empty string for each of the events.

![chrome_kri1BEb3fG](https://user-images.githubusercontent.com/8262173/74608351-dfa24100-50a5-11ea-97b3-51beb7548d50.png)

![chrome_exgsFEmOY1](https://user-images.githubusercontent.com/8262173/74608352-e03ad780-50a5-11ea-873c-42b13c5effff.png)

## Steps to Reproduce

1. Start OBS with `--remote-debugging-port=8081` so you have access to the debug tools.
2. Create a new scene in OBS Studio, add the Browser source, and point it at the default web page.
3. Go to `http://localhost:8081`. Run the following script in the console:
```
document.addEventListener("keypress", event => {
  console.log(event.code);
});
```
4. Right click on the Browser source and click Interact.
5. Hold down the A key after selecting the Interact window. This results in the wrong key code being logged in the debug console.

## Additional information

`event.keyCode` produces the correct values, but `event.code` is broken for both the `keypress` and `keydown` events.

WizardCM:
I can confirm this is an issue in browser sources.

Tested with: https://keycode.info

Findings:
* All keypresses are considered from the "general keys" event.location
* .key and .which are correct
* event.code returns undefined
* this works fine in browser panels

There are a few possibilities

1. Qt isn't correctly passing the events to Cef (may not be fixable by us)
2. Cef off-screen rendering doesn't track keypresses correctly (may not be fixable by us)
3. We're doing something wrong

WizardCM:
After doing some digging, this is primarily possibility 3, but not in the ways I expected.

[...]

PR #205, opened 2020-03-04
Slightly improve keyboard events in Interact

WizardCM:
### Description
Provides slightly better keyboard events to the webpage.

On all platforms, tries to provide the `KeyboardEvent.code` where possible
On Windows, this also provides an accurate `KeyboardEvent.key` and `KeyboardEvent.which`
On Linux, currently this provides largely useless key and which (but accurate `KeyboardEvent.code`). The big challenge here is [this is what the CEF tests](https://bitbucket.org/chromiumembedded/cef/src/e0974ea64d02545b594ae7d1e857a9d246c1f188/tests/cefclient/browser/browser_window_osr_gtk.cc?at=master&fileviewer=file-view-default#browser_window_osr_gtk.cc) (with mappings) look like. That's **a lot of code** for a few keypress events.

It's possible a mapping of ascii codes to JS .which values would do the trick for Linux.

### Motivation and Context
Partially improves #200 

### How Has This Been Tested?
Load http://keycode.info in a browser source and a browser panel. Note the differences when interacting with a panel (all keys show correctly) vs the Interact window for a source (largely empty or shows as F7).

Untested on macOS at this time.

### Types of changes
- Bug fix (non-breaking change which fixes an issue) -->

### Checklist:

### DIFF
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -295,12 +295,13 @@ void BrowserSource::SendKeyClick(const struct obs_key_event *event, bool key_up)
 	uint32_t modifiers = event->modifiers;
 	std::string text = event->text;
 	uint32_t native_vkey = event->native_vkey;
+	uint32_t native_scancode = event->native_scancode;
 
 	ExecuteOnBrowser(
 		[=](CefRefPtr<CefBrowser> cefBrowser) {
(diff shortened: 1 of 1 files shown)


WizardCM:
Fixed properly with #206

PR #206, opened 2020-03-05
obs-browser: Fix key mapping on linux

### LINKED ISSUE #200: Interact window provides incorrect codes on keypress and keydown events

Francexi:
### Description
The fix maps, only for Linux systems, XKB keycodes (that are provided to the SendKeyClick function) to the Javascript ones. The fix also add the native_scancode as value for the native_vkey, that seems to make the fix work also on Windows. (thanks to WizardCM for the collab)

### Motivation and Context
The keypress event was inusable on obs-browser. This seems to fully fix this

Fixes #200 

### How Has This Been Tested?
Tested on Slackware-current, using CEF build 3809. As long as I tested, all seems to work fine

### Types of changes

### Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,16 @@ if(APPLE)
 	endif()
 endif()
 
+if(UNIX AND NOT APPLE)
+	find_package(X11 REQUIRED)
+	if(NOT X11_FOUND)
+		message(STATUS "X11 Not found -- obs-browser plugin disabled.")
+		return()
+	endif()
+	include_directories("${X11_INCLUDE_DIR}")
+endif()
+
+
 configure_file(
 	"${CMAKE_CURRENT_SOURCE_DIR}/browser-config.h.in"
 	"${CMAKE_CURRENT_BINARY_DIR}/browser-config.h")
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -28,6 +28,10 @@
 #include <thread>
 #include <mutex>
 
+#ifdef __linux__
+#include "linux-keyboard-helpers.hpp"
+#endif
+
 #ifdef USE_QT_LOOP
 #include <QEventLoop>
 #include <QThread>
@@ -294,13 +298,19 @@ void BrowserSource::SendKeyClick(const struct obs_key_event *event, bool key_up)
 {
 	uint32_t modifiers = event->modifiers;
 	std::string text = event->text;
+#ifdef __linux__
+	uint32_t native_vkey = KeyboardCodeFromXKeysym(event->native_vkey);
+#else
 	uint32_t native_vkey = event->native_vkey;
+#endif
+	uint32_t native_scancode = event->native_scancode;
+	uint32_t native_modifiers = event->native_modifiers;
 
 	ExecuteOnBrowser(
 		[=](CefRefPtr<CefBrowser> cefBrowser) {
 			CefKeyEvent e;
 			e.windows_key_code = native_vkey;
-			e.native_key_code = 0;
+			e.native_key_code = native_scancode;
 
 			e.type = key_up ? KEYEVENT_KEYUP : KEYEVENT_RAWKEYDOWN;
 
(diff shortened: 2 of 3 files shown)


kkartaltepe:
Please add the appropriate library search in cmake along with adding the appropriate includes (but without linking since we do not actually need the library).

kkartaltepe:
You can see a good example of a simple find module at
https://github.com/obsproject/obs-studio/blob/master/cmake/Modules/FindX11_XCB.cmake

Adding a module searching pkg-config for `x11`(case matters) will get you the PKG_X11_LIBRARIES and PKG_X11_INCLUDE_DIR and follow the pattern in the above file for filling out the variables.

Like with findCef just drop it in the root of this package https://github.com/obsproject/obs-browser/blob/master/FindCEF.cmake

Then in the CMakeLists.txt you can simply `find_package(X11)` and use the X11_INCLUDE_DIR in an `include_directories($X11_INCLUDE_DIR)` in some Linux specific `if` block.

Francexi:
I don't know if this is right, I think is how it should work

kkartaltepe:
libX11-xcb is separate from libX11. The module for it is an example that is relatively simple. You should replace xcb references with plain x11 references. And change the header being searched for from the xcb header `X11/Xlib-xcb.h` to the X11 header you are using `X11/keysym.h`.

You can run `cmake` to test it. If running cmake with this commit didnt fail on your system you must already have an existing FindX11.cmake in your system module path that found it.

Francexi:
this seems to go well. I also passed the files through clang-format

CMakeLists.txt
```diff
@@ -116,7 +126,7 @@ set(obs-browser_HEADERS
 
 # only allow browser panels on win32 for now -- other operating systems
 # need more testing
-if(WIN32 AND BROWSER_PANEL_SUPPORT_ENABLED)
+if(BROWSER_PANEL_SUPPORT_ENABLED)
```
kkartaltepe: This seems unrelated to the PR. Did you mean to add this change?

Francexi: woops, that was unintended. I corrected it

obs-browser-source.cpp
```diff
@@ -290,17 +298,425 @@ void BrowserSource::SendFocus(bool focus)
...
+    case XF86XK_MonBrightnessDown:
+      return VKEY_BRIGHTNESS_DOWN;
+    case XF86XK_MonBrightnessUp:
+      return VKEY_BRIGHTNESS_UP;
+    case XF86XK_KbdBrightnessDown:
+      return VKEY_KBD_BRIGHTNESS_DOWN;
+    case XF86XK_KbdBrightnessUp:
+      return VKEY_KBD_BRIGHTNESS_UP;
+
+    // TODO(sad): some keycodes are still missing.
```
kkartaltepe: If you are going to copy and paste code from chromium at least document it so we know where to go to find updated versions.

For example this entire function is out of date with chromium head and noted that it only sort of works correctly on US layouts.
https://chromium.googlesource.com/chromium/src/+/lkgr/ui/events/keycodes/keyboard_code_conversion_x.cc

Francexi: well, that was the only mapping that I found that worked on both italian and US layouts (the two I had). If you want to suggest a modification in particular, can you do a pull request to my repo and I'll merge it

PR #210, opened 2020-03-24
Support Unicode in CSS via URI encoding

WizardCM:
### Description
Internally, atob() only supports ASCII, so we URI encode before base64 encoding.

### Motivation and Context
While it's not used often, CSS with Japanese characters didn't work properly. Now it does.

### How Has This Been Tested?
Add this to the Custom CSS field of a browser source:
```css
/* 24->OK, 25->NG */
yt-live-chat-renderer * {
  font-family: "ＭＳ ゴシック";
}
```
Then load the dev tools and view the resulting CSS block in the `<head>`.

### Types of changes

### Checklist:

### DIFF
--- a/browser-client.cpp
+++ b/browser-client.cpp
@@ -367,12 +367,13 @@ void BrowserClient::OnLoadEnd(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame> frame,
 	}
 
 	if (frame->IsMain()) {
-		std::string base64EncodedCSS = base64_encode(bs->css);
+		std::string uriEncodedCSS =
+			CefURIEncode(bs->css, false).ToString();
 
 		std::string script;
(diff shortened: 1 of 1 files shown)


notr1ch:
It seems redundant to use `atob` now, URI encoding should encode quotes.

WizardCM:
> It seems redundant to use `atob` now, URI encoding should encode quotes.

atob() had two purposes - quotes, and also newlines. Without it, newlines would cause the JS to become invalid.

notr1ch:
URI encoding should also handle newlines and any other control characters.

WizardCM:
I shall experiment and adjust it tomorrow.

WizardCM:
> URI encoding should also handle newlines and any other control characters.

You were correct, I've removed base64 entirely in favour of URI encoding.

PR #211, opened 2020-03-26
Set default CSS color to white

jpark37:
### Description
Black text on transparent is invisble with the OBS canvas.

### Motivation and Context
Could not see text at this URL: chrome://version

### How Has This Been Tested?
Can now see text at chrome://version

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-plugin.cpp
+++ b/obs-browser-plugin.cpp
@@ -104,6 +104,7 @@ bool QueueCEFTask(std::function<void()> task)
 static const char *default_css = "\
 body { \
 background-color: rgba(0, 0, 0, 0); \
+color: rgb(255, 255, 255); \
 margin: 0px auto; \
 overflow: hidden; \
 }";


Lange:
This is likely to have unexpected side effects for many web pages, due to the nature of the CSS cascade. Many people's graphics and alerts will have white text in certain places if they don't edit out this default style.

jpark37:
Bah, are there really that many alerts with inherited black text in them? I can't think of any off the top of my head, but I don't consume very many channels.

It would be nice if unstyled pages were readable by default, but go ahead and close this out if it's going to annoy too many people.

Lange:
You're probably right that the number of real-world cases this hits is relatively low, but I think it's too hard to predict. I'm personally ambivalent on this one, and defer to the support folks who would likely need to bear the brunt of any negative consequences caused by this change.

Fenrirthviti:
I agree that this is likely to cause more problems than solve them, but I don't have anything more than gut feeling to back that up. You can still view the text in interact mode, as a workaround. But, to be perfectly honest, I'm not sure that the custom CSS is actually helping as much these days, as the sites people are loading are already superseding our overrides on the body element, so it's as likely this change won't affect much.

example

WizardCM:
Other reasons why this is a bad idea, speaking with web development experience:
* OBS loads this CSS last, meaning it'll override anything the webpage developer intended (unless, of course, the developer is using `!important` for some reason)
* In the case of the chrome://version page, you'll notice that even if you remove all our custom CSS, the page is still transparent. That's because outside of the _webpage_, the "window" (the browser source) is transparent, where normally the container is white (see: browser panels)
* Most users leave the browser source CSS at default, which means even if the source was created before this change was made, this change will update the "old" CSS for existing sources, which can potentially change existing sources, and not in a beneficial way
* This change assumes the user wants to see the source on a black background (ie. no other sources). If this is purely to make it easier for yourself, create a Color Source (which defaults to white) and put it below the browser source.
* Generally, the less we override, the better. The website developer expects the web browser to behave like Google Chrome. Google Chrome does not override font colours.

jpark37:
- From loading other websites, it seemed like the custom CSS block was getting loaded first, but I checked with a test page, and you're right.
- I never realized OBS doesn't save settings unless they are modified, but I see that's true too.
- A Color Source is an ugly hack. I wish unstyled pages just worked out of the box.
- If we really want to match Chrome, an unstyled page should render white on black or black on white depending on Windows color scheme.

I'm going to leave this open a little white longer in case someone has a quick and easy idea to improve the unstyled page behavior, but I'm fine if someone comes along and closes out the issue too.

jpark37:
I don't care about this anymore, closing.

Issue #194, opened 2020-02-05
3974: CefAudioHandler is not a valid template type argument (in browser-client.cpp)


Xaymar:
When building obs-browser with CEF branch 3974 (or newer), an error can be observed as CefAudioHandler, ChannelLayout, GetAudioHandler and some other things are not part of CEF. These seem to be from patches that are not easily applied to a released CEF build and require custom CEF builds.

## Expected Behavior
obs-browser should build fine with CEF 3974.

## Current Behavior
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	C2923	'CefRefPtr': 'CefAudioHandler' is not a valid template type argument for parameter 'T' (compiling source file D:\Projects\Caffeine\obs-studio\plugins\obs-browser\obs-browser-source.cpp)	obs-browser	D:\Projects\Caffeine\obs-studio\plugins\obs-browser\browser-client.hpp	71	
Error	C2065	'CefAudioHandler': undeclared identifier	obs-browser	D:\Projects\Caffeine\obs-studio\plugins\obs-browser\browser-client.cpp	66	
Error	C2923	'CefRefPtr': 'CefAudioHandler' is not a valid template type argument for parameter 'T'	obs-browser	D:\Projects\Caffeine\obs-studio\plugins\obs-browser\browser-client.cpp	66	
Error	C2039	'GetAudioHandler': is not a member of 'BrowserClient'	obs-browser	D:\Projects\Caffeine\obs-studio\plugins\obs-browser\browser-client.cpp	66	
Error	C2065	'reroute_audio': undeclared identifier	obs-browser	D:\Projects\Caffeine\obs-studio\plugins\obs-browser\browser-client.cpp	68	
Error	C2355	'this': can only be referenced inside non-static member functions or non-static data member initializers	obs-browser	D:\Projects\Caffeine\obs-studio\plugins\obs-browser\browser-client.cpp	68	
Error	C2653	'CefAudioHandler': is not a class or namespace name	obs-browser	D:\Projects\Caffeine\obs-studio\plugins\obs-browser\browser-client.cpp	256	
Error	C2065	'ChannelLayout': undeclared identifier	obs-browser	D:\Projects\Caffeine\obs-studio\plugins\obs-browser\browser-client.cpp	256	
Error	C2146	syntax error: missing ')' before identifier 'cefLayout'	obs-browser	D:\Projects\Caffeine\obs-studio\plugins\obs-browser\browser-client.cpp	256	
[...]

PR #215, opened 2020-04-03
Remove audio routing through obs for cef >= 3865

ArisuOngaku:
### Description

Do not support audio routing through obs for CEF >= 3865.

### Motivation and Context

This is required because the api doesn't exist anymore. See https://bitbucket.org/chromiumembedded/cef/commits/1f893c76a97eaa88e3ff782b8e2e47dd0f5d27b2

Note that there is an open issue to reimplement this feature https://bitbucket.org/chromiumembedded/cef/issues/2755/reimplement-audio-mirroring-for

Related to #194 

### How Has This Been Tested?

On linux: built obs studio with cef-minimal 80.0.8 (3987) from the arch user repository (it launched and there was no crash).

### Types of changes

### Checklist:

### DIFF
--- a/browser-client.cpp
+++ b/browser-client.cpp
@@ -62,7 +62,7 @@ CefRefPtr<CefContextMenuHandler> BrowserClient::GetContextMenuHandler()
 	return this;
 }
 
-#if CHROME_VERSION_BUILD >= 3683
+#if CHROME_VERSION_BUILD >= 3683 && CHROME_VERSION_BUILD < 3865
 CefRefPtr<CefAudioHandler> BrowserClient::GetAudioHandler()
 {
 	return reroute_audio ? this : nullptr;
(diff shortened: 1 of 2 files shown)


WizardCM:
We do not recommend building OBS with CEF versions that don't support audio. For now, we suggest either 3770 on Windows and 3809 on other platforms.

ArisuOngaku:
@WizardCM This would enable (at least temporarily) building obs from source without the need of installing an old package version for arch user from the AUR (maybe there are other cases i.e. with gentoo?).

However, I would understand if the policy of this repo is to prefer making obs browser not compile when there is no audio routing support instead of taking the risk of unexpected behaviors (please close my PR if so).

WizardCM:
If building OBS from source, you can download the [required CEF version here](http://opensource.spotify.com/cefbuilds/index.html) (click "Show more builds", and choose "minimal"). It is its own CMake project, and once built you can add the CEF_ROOT_DIR to the OBS CMake project.

Appreciate your understanding. As soon as we've tested & are comfortable with newer versions of CEF, we'll update the codebase to correctly handle them.

Issue #226, opened 2020-05-04
[BUG] Scripts with type="module" are ignored


Rosuav:
## Platform

Operating system and version:
OBS Studio version: 25.0.2+ (built from git) on Debian Linux

## Expected Behavior
JavaScript in `<script type=module>` blocks should be executed.

## Current Behavior
JavaScript in such blocks is ignored.

## Steps to Reproduce
1. Create a browser source referencing https://rosuav.github.io/Ratigan/cutdown
2. The JS should create a list item (saying "visible"). This can be confirmed in a regular browser.

## Additional information
This script file is not using any module-specific features (eg imports), and should be executed just the same as if the type were omitted.

The issue does not seem to occur on Windows, but I have not confirmed this.

Fenrirthviti:
Can confirm it works fine on Windows 10.

Fenrirthviti:
Not sure if it helps anything, but on windows the event is triggered twice when showing/hiding the source for some reason: ![visibility](https://pub.rachni.com/img/obs64_2020-05-03_21-17-40.png)

Rosuav:
That's very curious. I've seen some other situations where the event double-fires, though (even in browsers), so I've no idea what the cause is. (If you want more info about the events themselves, try https://rosuav.github.io/Ratigan/comfy which establishes a socket connection to Twitch as well, allowing a bit of external interaction; it also gives more detailed info about the event timings and such.) That's tangential to this issue, which is that the script isn't even being run at all, if the type is module.

WizardCM:
The double-fire definitely isn't expected behaviour, but it's largely because we have to fake the event due to our current CEF build (and a few around it) having a broken visibility API. As long as it's reporting accurately, that should be fine.

If you're not seeing it fire, it's likely because you have Projector, Properties, Interact or Filters opened for the source. Due to the way OBS treats visibility, they are technically "always visible". Ideally you instead want to hook into the `obsSourceActiveChanged` event on the window/document instead. This will always give you a real status on whether the browser is visible in the stream/recording, regardless of previews visible. More info on the available [JS bindings here](https://github.com/obsproject/obs-browser#js-bindings).

Rosuav:
Curious. Why is it different on Linux and Windows?

WizardCM:
In a local build on Ubuntu with CEF 3809, the visibility API behaves identical to Windows, with the side effects mentioned in my earlier post. Which version of CEF are you building with?

Rosuav:
[...]

PR #228, opened 2020-05-08
Set a more usable default font size

Rosuav:
### Description
Set a default font size of 16px to ensure that text is readable when unstyled.

### Motivation and Context
The default font size in most browsers is 16px, but the default here appears sometimes to be 1px. This results in unreadable text (see eg the confusion in #226 which ultimately was caused by this), and is inconsistent.

### How Has This Been Tested?
Fired up several pages. Browser plugin was built against CEF 75.1.14+gc81164e+chromium-75.0.3770.100 - didn't test with other CEF versions. It might be that this patch only benefits older browser builds, but that's still worthwhile as 3770 is the one named in the OBS installation instructions.

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -168,6 +168,9 @@ bool BrowserSource::CreateBrowser()
 		cefBrowserSettings.windowless_frame_rate = fps;
 #endif
 
+		cefBrowserSettings.default_font_size = 16;
+		cefBrowserSettings.default_fixed_font_size = 16;
+
 #if ENABLE_LOCAL_FILE_URL_SCHEME
 		if (is_local) {
 			/* Disable web security for file:// URLs to allow


WizardCM:
In testing, this is exclusive to 3770 and only on Linux as far as we could tell. Is it common for apps using CEF to manually set the font size?

Rosuav:
I'm not saying the *bug* is exclusive to 3770/Linux - just that that's all I have access to test it on. I'm currently trying to build other CEFs for testing, but that's its own job and may or may not succeed.

Rosuav:
No idea how common it is for apps to manually set the font size, but given that I'm setting it to the most commonly-used default, it's going to do nothing if it wasn't a problem.

WizardCM:
No I know, I mean when we were preparing 25.x builds with the browser source, we ran into the same issue, and the reason we ended up jumping to 3809 on our official build is because the issue didn't occur there. We originally thought it was a rendering bug, but turned out to be font-size related.

Personally I think this change is fine if other browsers also use it as default, as long as we can verify that.

Rosuav:
Actually, can confirm that this does not happen on 3809, so this is a fix to ensure that the version listed in the installation instructions does work. Alternatively, changing the installation instructions (and including a fully-built CEF for download) would resolve this issue.

aerogus:
Great idea this default font size. It may resolve the bug I have with "em" or "rem" units in my CSS. I had to switch to "px" units.

Rosuav:
Yes, it should resolve that (based on my testing).

PR #233, opened 2020-07-23
Update macOS minimum version to 10.13

RytoEX:
### Description

Update the minimum macOS version targeted/supported to macOS 10.13.

### Motivation and Context

In line with obsproject/obs-deps#16, since we no longer actively support macOS older than 10.13, let's update the minimum macOS version targeted in our code and build scripts.

### How Has This Been Tested?

I have not tested this.  Someone with able to build and test on macOS 10.13+ will need to build and test this.

### Types of changes

### Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -232,7 +232,7 @@ if (APPLE AND NOT BROWSER_LEGACY)
     		MACOSX_BUNDLE_INFO_PLIST ${_helper_info_plist}
     		OUTPUT_NAME ${_helper_output_name}
             FOLDER "plugins/obs-browser/helpers/"
-			COMPILE_FLAGS "-mmacosx-version-min=10.8"
+			COMPILE_FLAGS "-mmacosx-version-min=10.13"
     	)
 
         if(XCODE)
@@ -271,6 +271,10 @@ else()
     if(APPLE)
         target_compile_definitions(obs-browser PRIVATE BROWSER_LEGACY)
     endif()
+endif()
+if(APPLE)
+	set_target_properties(obs-browser-page PROPERTIES
+		COMPILE_FLAGS "-mmacosx-version-min=10.13")
 endif()
 
 if (WIN32)


RytoEX:
It seems that trying to build the CEF wrapper (libcef_dll_wrapper) with 10.13 as a target causes some build failures.  These failures showed up in my obs-studio fork when trying to build on CI.  The failures appear to be due to CEF using functions deprecated in 10.12.  The failures may be able to be suppressed by omitting -Werror and/or -Wdeprecated-declarations, but I'm not sure if that's okay here.  On the other hand, it's third-party code that we don't modify, so we aren't responsible for errors and deprecation warnings there.

I can try removing those warning flags specifically for building libcef_dll_wrapper, but I'd appreciate some feedback on that action.

WizardCM:
The main reason we have a 10.13 minimum requirement is due to a Qt update, rather than outdated code on our side. I think it's fine if the browser panels are built with an "outdated" minimum version as it won't affect the user experience.

When we next update CEF, we can make sure to update the minimum macOS version to match. We could create & pin an Issue for this so that we don't forget.

RytoEX:
Unfortunately, I think the deprecation errors are going to be a problem in the future.  The same code that CI is failing on is currently in CEF master, so a CEF update won't fix this.

The hope behind submitting updates for obs-deps, obs-browser, and obs-studio is that by having everything targeting the same minimum macOS we can gain some benefits by not having the build tools and compiler think we need support for 10.8.

WizardCM:
In that case, would it be worth checking the CEF Issues list, and if it's not documented, to raise it ourselves?

RytoEX:
> In that case, would it be worth checking the CEF Issues list, and if it's not documented, to raise it ourselves?

It's probably not a bad idea.  The closest I could find in their PR/issues lists is this:  https://bitbucket.org/chromiumembedded/cef/pull-requests/320/update-mac-target-sdk-to-1010/diff

Not sure what the best approach is for bringing this up there.  File a new issue?  Any insight @pkviet ?

RytoEX:
It looks like CEF itself has [an open PR](https://bitbucket.org/chromiumembedded/cef/pull-requests/303/cefclient-port-to-gtk3/diff) where Marshall suggested adding `-Wno-deprecated-declarations` to their `cmake/cef_variables.cmake` file to avoid build errors.  I've seen other projects use `-Wno-deprecated-declarations` and/or `-Wno-error=deprecated-declarations` to prevent deprecation errors/warnings from dependencies, so I think this is relatively safe to do for our purposes here.

As far as I know, `-Wno-error=deprecated-declarations` turns errors for deprecated function calls into warnings, so it just prevents the build from failing, but the warnings may still show up in build logs.  If we wanted to also suppress the warnings, we could probably also use `-Wno-deprecated-declarations` just for the libcef_dll_wrapper build.

PatTheMav:
@RytoEX yeah that's exactly what I just added to CI for macOS 10.13 "compatibility".

RytoEX:
> 
> 
> @RytoEX yeah that's exactly what I just added to CI for macOS 10.13 "compatibility".

@PatTheMav 
Hmm.  I guess I never submitted my branch as a PR for some reason.  I probably just forgot.
https://github.com/RytoEX/obs-studio/tree/update-macos-min
https://github.com/RytoEX/obs-studio/commit/d829a87374c186137cb023deb37ff14b533290cd

WizardCM:
So what's the status on this one? Likely best to save it for when we update CEF?

RytoEX:
> When we next update CEF, we can make sure to update the minimum macOS version to match. We could create & pin an Issue for this so that we don't forget.

> So what's the status on this one? Likely best to save it for when we update CEF?

I somehow didn't see the notification for this.  As we are updating CEF, should we merge this?

PatTheMav:
Theoretically we should be able to merge this already, as the dll wrapper is built for macOS 10.13 anyway.

WizardCM:
Please rebase this branch, and then I'll merge this.

RytoEX:
Rebased and targeted 10.13 as the minimum in both places it was specified.  It was previously only specified once, but @PatTheMav has made some additions since this was originally PR'd.

PR #248, opened 2020-11-25
Fix Xcode linking issues introduced by Cmake 3.19.0

PatTheMav:
### Description
Fixes Xcode build failing when created with Cmake 3.19.0+ and updates minimum supported macOS version to align with the main project.

### Motivation and Context
Cmake 3.19.0+ switched to Xcode's modern build system, which also changed the way frameworks are linked/added to the project. Whereas https://gitlab.kitware.com/cmake/cmake/-/commit/ce2dee9e5bae37c8117087bb83add075c3c123b4 added support to escape the framework *path*, it doesn't escape the framework *name* which in the case of CEF contains spaces and breaks the `LDFLAG` setting.

### How Has This Been Tested?
* OBS successfully configured and compiled with Xcode
* OBS successfully configured and compiled from CLI

### Types of changes

### Checklist:

### DIFF
--- a/FindCEF.cmake
+++ b/FindCEF.cmake
@@ -67,9 +67,19 @@ if(WIN32)
 				debug ${CEFWRAPPER_LIBRARY_DEBUG})
 	endif()
 else()
-	set(CEF_LIBRARIES
-			${CEF_LIBRARY}
-			${CEFWRAPPER_LIBRARY})
+	# Fixes cmake 3.19.0 commit that added support for modern Xcode build system, but "forgot"
+	# to also escape framework names themselves in addition to the framework path:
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -177,7 +177,7 @@ set_target_properties(obs-browser-page PROPERTIES FOLDER "plugins/obs-browser")
 
 if (APPLE)
 	set_target_properties(obs-browser-page PROPERTIES
-		COMPILE_FLAGS "-mmacosx-version-min=10.8")
+		COMPILE_FLAGS "-mmacosx-version-min=10.13")
```
WizardCM, line 180: Could you remove this change so that we can retain the discussions in #233?

PatTheMav: Removed and force-pushed an update.

PR #249, opened 2020-11-26
obs-browser: Update to new audio api

pkviet:
### Description
This updates the audio functions due to api changes in cef & chromium.
(new AudioService API used for cef >= 4103)
Compatibility with cef 3770 is kept for easier fallback.
Minor version of obs-browser updated to 2.9.0.
The audio mixing has been removed since the new API does it automatically.
Audio lines are no longer used.
NB: natives_blob.bin is no longer incliuded in cef so a check is added to cmake.

### Motivation and Context
Allows to update to newer cef versions.

### How Has This Been Tested?
Checked that audio worked on cef 4183.
Checked that audio with several streams was indeed auto-mixed without obs audio lines.

### Types of changes
- New feature (non-breaking change which adds functionality) -->
- Performance enhancement (non-breaking change which improves efficiency)

### Checklist:


WizardCM:
I have verified the audio changes work with CEF 4183, and also tested the 3770 fallback. As a precaution, I also tested 3809, which we use on Linux. Everything works as it should.

PR #252, opened 2020-11-30
Update CEF to support multi-process structure required for macOS

PatTheMav:
### Description
This PR adds the necessary build-system changes for modern CEF versions running on macOS and re-enables features.

### Motivation and Context
To ensure compatibility with macOS Catalina and beyond Chromium requires to be run via multiple processes on macOS with the single-process model not being supported anymore.

As such Chromium requires 3 additional helper apps to exist (GPU, Plugin and Renderer) and also the main helper to be packaged in a macOS app bundle. The helpers also require specific entitlements to allow JIT compilation.

This PR adds the necessary changes to cmake to build these helpers. Changes to OBS packaging are also required but are addressed in a PR on the main repo.

*Note:* The helpers for all intents and purposes replace `obs-browser-page` on macOS. Also the option to deploy the browser is now assumed by default as CEF will not run when not linked/placed in a macOS app bundle.

### How Has This Been Tested?
* OBS built with CEF and tested with multiple browser sources.
* H.264 and VP9 video playback tested in browser sources

### Types of changes

### Checklist:


PatTheMav:
This requires our planned CEF 4183 update to work.

PatTheMav:
After a quick chat with @pkviet we decided to add a `BROWSER_LEGACY` flag that will allow to easily fall back on 3770-era behaviour.

I'll also investigate adding a similar flag to the build scripts and Github workflow to allow a fallback if excrements start hitting a blade-based appliance.

PatTheMav:
Added the `BROWSER_LEGACY` flag, it's automatically set by the updated scripts in https://github.com/obsproject/obs-studio/pull/3805 if `CEF_BUILD_VERSION=3770` or lower is defined.

WizardCM:
@pkviet have you tested this branch locally? If so, please mark it as approved.

PatTheMav:
There are still a few fixes missing that I noticed while dogfooding this, will push an update tomorrow/later.

PatTheMav:
After defeating `clang_format` once more, I successfully pushed the small changes I made recently. With those:

* CEF is loaded at runtime on macOS as suggested by the CEF developers, `CefScopedLibraryLoader` is used, which requires the helper-app infrastructure to work
* The Win32-only call `CefEnableHighDPISupport` has been disabled for other platforms
* Cmake files have been adapted to not link CEF directly on macOS

The second commit adds the `BROWSER_LEGACY` flag to all those changes, "undoing" them if set.

The third commit adds a little change required to make texture sharing work with CEF 4183 later.

obs-browser-plugin.cpp
```diff
@@ -224,6 +224,13 @@ static void BrowserInit(void)
 	path += ".exe";
 	CefMainArgs args;
 #else
+#if defined(__APPLE__) && !defined(BROWSER_DEPLOY)
+	/* Load CEF at runtime as required on macOS */
+	CefScopedLibraryLoader library_loader;
+	if (!library_loader.LoadInMain())
+		return;
```
PatTheMav, line 231: @WizardCM This is actually not so great, because the early return isn't considered by the plugin infrastructure. If you have a suggestion how to handle the case of CEF not being able to load the library and abort/quit gracefully, I can adapt that.

WizardCM: Just for my reference, is this a blocker in terms of merging?

PatTheMav: It's just uncool as you get a blank browser source and the log is spammed with "cannot load" messages, but if the CEF library is in the bundle, it should load it just fine.

obs-browser-plugin.cpp
```diff
@@ -579,9 +588,9 @@ bool obs_module_load(void)
 
 	os_event_init(&cef_started_event, OS_EVENT_TYPE_MANUAL);
 
-	CefEnableHighDPISupport();
-
 #ifdef _WIN32
+	/* CefEnableHighDPISupport doesn't do anything on OS other than Windows. Would also crash macOS at this point as CEF is not directly linked */
```
PatTheMav, line 592: Felt cute, might break up later - also `clang_format` didn't mind.. 😬

PatTheMav:
Just pushed a new iteration of the commits - main changes:

* `SHARED_TEXTURE_SUPPORT_ENABLED` is now a simple define set by `cmake`, no futzing around with a pre-configured header file now (it's also not really that experimental anymore)
* I've disabled keychain access for the plugin; we disable sandboxing and other security features and thus should not get access to the device keychain on macOS
* As before macOS now uses the CefLibraryLoader and does not link to CEF directly anymore
* `CefExecuteProcess` is only executed on Windows, it should not be called on macOS due to its helper-process architecture
* `SendExternalBeginFrame` doesn't seem to work properly on macOS, so I made it a Windows-only feature for now
* As an alternative the canvas framerate is now set for browser sources by default, but is not adjusted automatically
* The texture created from the shared `IOSurface` does not have pre-multiplied alpha, so `OBS_EFFECT_DEFAULT_RECT` is used if texture sharing is enabled
* As before the second commit introduces the flag `BROWSER_LEGACY` which, combined with CEF v3770, allows building a "last-known good" version of the plugin

I'll give this a whirl with different build configs tomorrow, but it's looking good so far. With `SendExternalBeginFrame` disabled I was not able to get the GPU Helper to crash and got solid 60fps in FishGL with 200+ fish and somewhere between 4-8% CPU and ~20% GPU usage (GPU Helper + OBS itself).

PR #254, opened 2020-12-08
Add Linux browser panel support [CEF 4280 / Chromium 87]

cg2121:
### TODO
- [x] Needs more testing (tested on Ubuntu 20.04 with QT 5.12.8)
- [x] Un-hardcode obs-browser.so lookup
- [x] Fix random crashes that sometimes happen

### Description
Add Linux browser panel support.

Screenshot from 2020-12-08 14-39-41

### Motivation and Context
Bringing the browser plugin to parity with Windows.

### How Has This Been Tested?
Needs testing on other OSs to make sure it doesn't break anything.

### Types of changes

### Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,9 +137,7 @@ if (APPLE)
 		)
 endif()
 
-# only allow browser panels on win32 for now -- other operating systems
-# need more testing
-if((WIN32 OR APPLE) AND BROWSER_PANEL_SUPPORT_ENABLED)
+if(BROWSER_PANEL_SUPPORT_ENABLED)
 	list(APPEND obs-browser_SOURCES
 		panel/browser-panel.cpp
 		panel/browser-panel-client.cpp
--- a/panel/browser-panel.cpp
+++ b/panel/browser-panel.cpp
@@ -20,6 +20,10 @@
 #include <util/base.h>
 #include <thread>
 
+#if !defined(_WIN32) && !defined(__APPLE__)
+#include <X11/Xlib.h>
+#endif
+
 extern bool QueueCEFTask(std::function<void()> task);
 extern "C" void obs_browser_initialize(void);
 extern os_event_t *cef_started_event;
@@ -86,6 +90,7 @@ struct QCefCookieManagerInternal : QCefCookieManager {
 		BPtr<char> rpath = obs_module_config_path(storage_path.c_str());
 		if (os_mkdirs(rpath.Get()) == MKDIR_ERROR)
 			throw "Failed to create cookie directory";
+
 		BPtr<char> path = os_get_abs_path_ptr(rpath.Get());
 
 #if CHROME_VERSION_BUILD < 3770
@@ -171,6 +176,9 @@ QCefWidgetInternal::QCefWidgetInternal(QWidget *parent, const std::string &url_,
 	setAttribute(Qt::WA_NativeWindow);
 
 	setFocusPolicy(Qt::ClickFocus);
+
+	window = new QWindow();
+	window->setFlags(Qt::FramelessWindowHint);
 }
 
 QCefWidgetInternal::~QCefWidgetInternal()
@@ -231,28 +239,28 @@ void QCefWidgetInternal::closeBrowser()
 
 void QCefWidgetInternal::Init()
 {
-	WId id = winId();
+	WId handle = window->winId();
 
-	bool success = QueueCEFTask([this, id]() {
+	QSize size = this->size();
+#ifdef SUPPORTS_FRACTIONAL_SCALING
+	size *= devicePixelRatioF();
+#else
+	size *= devicePixelRatio();
+#endif
+
+	bool success = QueueCEFTask([this, handle, size]() {
 		CefWindowInfo windowInfo;
 
 		/* Make sure Init isn't called more than once. */
 		if (cefBrowser)
 			return;
 
-		QSize size = this->size();
 #ifdef _WIN32
-#ifdef SUPPORTS_FRACTIONAL_SCALING
-		size *= devicePixelRatioF();
-#elif
-		size *= devicePixelRatio();
-#endif
 		RECT rc = {0, 0, size.width(), size.height()};
-		windowInfo.SetAsChild((HWND)id, rc);
-#elif __APPLE__
-		windowInfo.SetAsChild((CefWindowHandle)id, 0, 0, size.width(),
-				      size.height());
+#else
+		CefRect rc = {0, 0, size.width(), size.height()};
 #endif
+		windowInfo.SetAsChild((CefWindowHandle)handle, rc);
 
 		CefRefPtr<QCefBrowserClient> browserClient =
 			new QCefBrowserClient(this, script, allowAllPopups_);
@@ -264,13 +272,19 @@ void QCefWidgetInternal::Init()
 			CefRefPtr<CefDictionaryValue>(),
 #endif
 			rqc);
-#ifdef _WIN32
-		Resize();
-#endif
 	});
 
-	if (success)
+	if (success) {
 		timer.stop();
+
+		if (!container) {
+			container =
+				QWidget::createWindowContainer(window, this);
+			container->show();
+		}
+
+		Resize();
+	}
 }
 
 void QCefWidgetInternal::resizeEvent(QResizeEvent *event)
@@ -281,23 +295,47 @@ void QCefWidgetInternal::resizeEvent(QResizeEvent *event)
 
 void QCefWidgetInternal::Resize()
 {
-#ifdef _WIN32
 #ifdef SUPPORTS_FRACTIONAL_SCALING
 	QSize size = this->size() * devicePixelRatioF();
-#elif
+#else
 	QSize size = this->size() * devicePixelRatio();
 #endif
 
-	QueueCEFTask([this, size]() {
+	bool success = QueueCEFTask([this, size]() {
 		if (!cefBrowser)
 			return;
-		HWND hwnd = cefBrowser->GetHost()->GetWindowHandle();
-		SetWindowPos(hwnd, nullptr, 0, 0, size.width(), size.height(),
+
+		CefWindowHandle handle =
+			cefBrowser->GetHost()->GetWindowHandle();
+
+		if (!handle)
+			return;
+
+#ifdef _WIN32
+		SetWindowPos((HWND)handle, nullptr, 0, 0, size.width(),
+			     size.height(),
 			     SWP_NOMOVE | SWP_NOOWNERZORDER | SWP_NOZORDER);
-		SendMessage(hwnd, WM_SIZE, 0,
+		SendMessage((HWND)handle, WM_SIZE, 0,
 			    MAKELPARAM(size.width(), size.height()));
-	});
+#elif __APPLE__
+#else
+		Display *xDisplay = cef_get_xdisplay();
+
+		if (!xDisplay)
+			return;
+
+		XWindowChanges changes = {0};
+		changes.x = 0;
+		changes.y = 0;
+		changes.width = size.width();
+		changes.height = size.height();
+		XConfigureWindow(xDisplay, (Window)handle,
+				 CWX | CWY | CWHeight | CWWidth, &changes);
 #endif
+	});
+
+	if (success && container)
+		container->resize(size.width(), size.height());
 }
 
 void QCefWidgetInternal::showEvent(QShowEvent *event)
--- a/panel/browser-panel.hpp
+++ b/panel/browser-panel.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <obs-module.h>
 #include <util/platform.h>
 #include <util/util.hpp>
 #include <QWidget>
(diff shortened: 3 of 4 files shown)


CMakeLists.txt
```diff
@@ -126,7 +126,7 @@ set(obs-browser_HEADERS
 
 # only allow browser panels on win32 for now -- other operating systems
 # need more testing
-if(WIN32 AND BROWSER_PANEL_SUPPORT_ENABLED)
+if(BROWSER_PANEL_SUPPORT_ENABLED)
```
RytoEX, line 129: Might as well change the comment to reflect this isn't just WIN32 anymore.

panel/browser-panel.cpp
```diff
@@ -16,6 +16,10 @@
 #include <util/base.h>
 #include <thread>
 
+#if !defined(_WIN32) && !define(__APPLE__)
```
RytoEX, line 19: Shouldn't this be `#if !defined(_WIN32) && !defined(__APPLE__)`?
Note that you use `define` instead of `defined` for the second condition.

panel/browser-panel.cpp
```diff
@@ -253,9 +256,8 @@ void QCefWidgetInternal::Init()
 			CefRefPtr<CefDictionaryValue>(),
 #endif
 			rqc);
-#ifdef _WIN32
+
```
RytoEX, line 259: Probably don't need this whitespace addition, unless the visual separation is truly needed.

panel/browser-panel.cpp
```diff
@@ -280,24 +281,42 @@ void QCefWidgetInternal::Resize()
...
+#ifdef _WIN32
 		HWND hwnd = cefBrowser->GetHost()->GetWindowHandle();
 		SetWindowPos(hwnd, nullptr, 0, 0, size.width(), size.height(),
 			     SWP_NOMOVE | SWP_NOOWNERZORDER | SWP_NOZORDER);
 		SendMessage(hwnd, WM_SIZE, 0,
 			    MAKELPARAM(size.width(), size.height()));
-	});
+#elif __APPLE__
+#else
+		auto browser_win = cefBrowser->GetHost()->GetWindowHandle();
```
RytoEX, line 293: I think in C++ files we use camelCase for variables.  I'd probably use `browserWindow` for this.

RytoEX (changes requested):
Just a few notes based on a visual readthrough.  I imagine between this and #197, one will need rebased when the other is merged.

CMakeLists.txt
cg2121: Fixed.

cg2121:
All requested fixes in the review are now fixed.

panel/browser-panel.cpp
```diff
@@ -237,10 +239,11 @@ void QCefWidgetInternal::Init()
 		size *= devicePixelRatio();
 #endif
 		RECT rc = {0, 0, size.width(), size.height()};
-		windowInfo.SetAsChild((HWND)id, rc);
-#elif __APPLE__
-		windowInfo.SetAsChild((CefWindowHandle)id, 0, 0, size.width(),
-				      size.height());
+		windowInfo.SetAsChild((HWND)this->window->winId(), rc);
+#else
+		CefRect rc = {0, 0, size.width(), size.height()};
```
RytoEX, line 244: What is the difference between RECT and CefRect?  Is this a platform specific difference?

panel/browser-panel.cpp
```diff
@@ -270,7 +271,6 @@ void QCefWidgetInternal::resizeEvent(QResizeEvent *event)
 
 void QCefWidgetInternal::Resize()
 {
-#ifdef _WIN32
```
RytoEX, line 273: If you're removing the `#ifdef _WIN32` here for `SUPPORTS_FRACTIONAL_SCALING`, doesn't that mean the instance on line 233/235 should also be moved/removed?

RytoEX (changes requested):
A couple more things.

WizardCM:
Ubuntu 20.04:

> error: X Error: BadWindow (invalid Window parameter), Major opcode: XI_BadMode (invalid Mode parameter), Minor opcode: 46, Serial: 250

When I create a custom browser dock.

> terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Aborted (core dumped)

When I try to open the Service Integration window (specifically to connect Twitch).

Browser sources work fine.

WizardCM:
OK, after a bit of debugging myself, the cause for the second error above is

https://github.com/obsproject/obs-browser/blob/ce8f6f8269275c4fd672ef8ad25225a6b9418b78/panel/browser-panel.cpp#L99

Essentially, `path.Get()` returns `null`.

https://github.com/obsproject/obs-browser/blob/ce8f6f8269275c4fd672ef8ad25225a6b9418b78/panel/browser-panel.cpp#L82-L83

 I *think* this is because the value it's based on, `rpath`, ends up being an absolute path rather than an expected relative path, so the conversion fails.

panel/browser-panel.hpp
```diff
@@ -72,7 +72,8 @@ static inline QCef *obs_browser_init_panel(void)
 #ifdef _WIN32
 	void *lib = os_dlopen("obs-browser");
 #else
-	void *lib = os_dlopen("../obs-plugins/obs-browser");
+	// Need to un-hardcode this, as I couldn't get it to work otherwise
+	void *lib = os_dlopen("/usr/lib/obs-plugins/obs-browser.so");
```
WizardCM, line 76: Would @tbodt's solution on macOS work here too?

https://github.com/obsproject/obs-browser/pull/197/files#diff-69dd8c025b947c895c80b4705d56a61d880c8841834614a0643b6cb727e4cd19R98-R99

panel/browser-panel.cpp
WizardCM: My guess based on the documentation is that CefRect is a platform-agnostic method.

https://magpcss.org/ceforum/apidocs3/projects/(default)/CefWindowInfo.html#SetAsChild(cef_window_handle_t,RECT)

cg2121: Changed to CefRect, still needs confirmation it still compiles on Windows.

panel/browser-panel.cpp
cg2121: Fixed.

panel/browser-panel.hpp
cg2121: Didn't work for me.

panel/browser-panel.cpp
RytoEX: My hope is that if we use CefRect everywhere, it's more uniform and less code in ifdefs.

cg2121:
Service integration should be working with the latest push.

WizardCM:
> error: X Error: BadWindow (invalid Window parameter), Major opcode: XI_BadMode (invalid Mode parameter), Minor opcode: 46, Serial: 250

I've spent a few hours trying to bypass this error. No luck. If I remove the `windowInfo.SetAsChild` call, a secondary window does open, but that's not the goal.

I tried a number of variations, including using `this->effectiveWindowId()` or manually setting `windowInfo.parent_window`. Everything would return the same error.

The only way I was able to change the behaviour is using `XConfigureWindow` and `XReparentWindow` after `CreateBrowserSync`, which leaves an extra empty black window which then freaks out and cannot be closed. The panel itself renders correctly but cannot be interacted with or resized.

I don't know X well enough to be able to do this using X functions correctly.

cg2121:
@WizardCM could you check if it works better for you now? I removed all of the QT functions from the CEF threads, and it seems it is more stable. It also seemed to work better on the latest CEF (CEF 4280).

WizardCM:
@cg2121 Nice work! I can confirm that they now correctly appear inside the dock (and within the Service Integration login container).

As a side note, I'm using CEF 4183 as that's what we hope to upgrade to.

However, while the child windows are correctly added, the following things don't work:

1. I cannot click within the webpage to interact with it (either within a dock, or service integration login)
2. Resizing the dock/login doesn't resize the webpage, and instead shows the empty container
3. When closing service integration login, both `BadWindow: 25` and `BadWindow: BadName` occur (if I resize before closing, I also get `BadWindow: BadColor` for every resize event)
4. After triggering 3, on shutdown I get `Trace/breakpoint trap (core dumped)`

cg2121:
@WizardCM the crash on exit appears to only happen on CEF 4183, but resizing and clicking work fine for me.

WizardCM:
@cg2121 Hmm, in that case I might try bumping to 4240 or 4280 for my testing.

WizardCM:
@cg2121 I can confirm that 4240 doesn't have any change in behaviour at all, but 4280 fixes the crash on shutdown and the two BadWindow errors when closing service integration login.

The rest of the issues remain.

RytoEX:
> Needs testing on other OSs to make sure it doesn't break anything.

What OS did you test on so far?  Can you please add that information to the original post?

cg2121:
I've tested it so far on Ubuntu 20.04 with QT 5.12.8.

phedders:
On Fri, 2020-12-11 at 11:44 -0800, Clayton Groeneveld wrote:
> I've tested it so far on Ubuntu 20.04 with QT 5.12.8.


I've been testing on Debian Unstable
- QT 5.15
- cef_binary_87.1.11+g8bb7705+chromium-87.0.4280.66_linux64
- obs-studio HEAD
- obs-browser remotes/lbp/linux-browser-panels 24c371a

So far everything I have thrown at it has worked.
- enable gpu works brilliantly.
- docks work
- browser windows work perfect - no errors or crashes so far
- ninja and other GPU intensives work really well with --enable-gpu

For my builds I'm using this to make starting with Accell easy.

> commit 3cdddbb06910ffa6705041340b80c8f556114d7c
> Author: Paul Hedderly <paul+github@.mjr.org>
> Date:   Mon Dec 7 22:10:03 2020 +0000
> 
>     prh: change desktop file for accell
> 
> diff --git a/UI/xdg-data/com.obsproject.Studio.desktop b/UI/xdg-
> data/com.obsproject.Studio.desktop
> index 4b753a4f5..cdf466e95 100644
> --- a/UI/xdg-data/com.obsproject.Studio.desktop
> +++ b/UI/xdg-data/com.obsproject.Studio.desktop
> @@ -6,7 +6,7 @@ GenericName[fr]=Logiciel d'enregistrement/diffusion
>  Comment=Free and Open Source Streaming/Recording Software
>  Comment[fr]=Logiciel libre d'enregistrement et de diffusion sur
> Internet
>  Comment[ru]=Бесплатная программа с открытым кодом для
> записи/трансляции видео
> -Exec=obs
> +Exec=obs --enable-gpu --enable-media-stream
>  Icon=com.obsproject.Studio
>  Terminal=false
>  Type=Application

cg2121:
@WizardCM by chance, are you using a VM to test it?

panel/browser-panel.cpp
cg2121: Nope. It has to use RECT on Windows.

cg2121:
I've now tested on Windows, resizing doesn't quite work as expected, though.

WizardCM:
> @WizardCM by chance, are you using a VM to test it?

Correct, I'm using my PXE boot in a Hyper-V VM. For some reason it has had trouble booting on real hardware recently, I hope to troubleshoot that this weekend.

panel/browser-panel.cpp
RytoEX: That's unfortunate.  Oh well.

cg2121:
The obs-browser library lookup is now un-hardcoded. Requires https://github.com/obsproject/obs-studio/pull/3871 to work.

panel/browser-panel.hpp
cg2121: This library lookup is fixed now.

eNBeWe:
I just built this code on Ubuntu 20.04:
- With the CEF version from the obs-studio documentation (3770) the browser contents are super tiny (approx 1% zoom level)
- With CEF 4280 the first tests look quite promising
- I would like to test the twitch integrations, but can't authenticate to twitch with the locally built version.

cg2121:
I've fixed the Windows resize issue. All I had to do was to set the window container as frameless.

WizardCM:
I can confirm that the lib lookup function works perfectly, even with nonstandard locations for OBS.

**Heads up:** the issues I was running into with resizing and interaction are caused by the `USE_QT_LOOP` CMake flag, which is designed to fix conflicts between Qt, CEF & GTK. I have this enabled to emulate production builds as much as possible. Getting browser panels to work properly with this flag will require a fair bit of rework of the panel code, but should solve a lot of issues.

> * With the CEF version from the obs-studio documentation (3770) the browser contents are super tiny (approx 1% zoom level)

@eNBeWe This is why we use 3809 on Linux. Basically, 3770 doesn't seem to set the default font size, so all text is tiny (everything else should be fine). Alternatively, [add this commit](https://github.com/obsproject/obs-browser/pull/228).

> * The embedded browser is unable to play twitch streams. I would like to use it to monitor the outgoing stream, which doesn't work because of this.

This is likely because non-free codecs are not enabled in Spotify builds.

panel/browser-panel.hpp
```diff
@@ -67,13 +68,19 @@ struct QCef {
 					 QObject *obj) = 0;
 };
 
+static void *get_browser_lib()
+{
+	obs_module_t *browserModule = obs_get_module_by_name("obs-browser");
```
tbodt, line 73: I just tried to port this over to the mac branch and got a "use of undeclared identifier": there apparently is no such function. Is this a typo?

tbodt: Ah, https://github.com/obsproject/obs-studio/pull/3871.

cg2121:
With the newer CEF versions (tested with 4280), the USE_QT_LOOP on Linux doesn't seem necessary. From what I can tell, the crashes don't happen anymore with newer CEF versions, as I haven't been able to reproduce them.

cg2121:
If that turns out to be the case, then this PR is ready to go.

cg2121:
Confirmed with another developer that they aren't getting the crashes anymore either, so as is this can be merged without problems, at least for Linux.

PatTheMav:
> OK, after a bit of debugging myself, the cause for the second error above is
> 
> https://github.com/obsproject/obs-browser/blob/ce8f6f8269275c4fd672ef8ad25225a6b9418b78/panel/browser-panel.cpp#L99
> 
> Essentially, `path.Get()` returns `null`.
> 
> https://github.com/obsproject/obs-browser/blob/ce8f6f8269275c4fd672ef8ad25225a6b9418b78/panel/browser-panel.cpp#L82-L83
> 
> I _think_ this is because the value it's based on, `rpath`, ends up being an absolute path rather than an expected relative path, so the conversion fails.

@tbodt added a fix for this in https://github.com/obsproject/obs-browser/pull/197, so maybe adapt that here as well?

I guess merging https://github.com/obsproject/obs-browser/pull/197 first makes sense, then the PRs for "modern" CEF versions required for macOS (because that would include CEF 4280 as well), and finally this PR?

cg2121:
I've added the fix from #197. It also works on Linux without problems.

flexiondotorg:
Thank for working on this! I've been testing on Ubuntu 20.04 with OBS 26.0.1 with the following patches applied:

  * https://github.com/obsproject/obs-studio/pull/3871
  * https://github.com/obsproject/obs-browser/pull/254

I've also dropped `USE_QT_LOOP=TRUE` from my build. Working great, no issues encountered.

LiamDawe:
Tested the Snap package noted by @flexiondotorg, looking good over here on Arch Linux. Docked my Twitch chat window without issues.

WizardCM:
@cg2121 Now that macOS panel support has been merged, please rebase your branch.


In the meantime, I'm going to try and get `USE_QT_LOOP` on linux to behave, just so we don't destroy existing configurations if possible. It won't be a requirement for this PR to be merged, but it's something I think would be worth doing.

cg2121:
It has been updated to the latest master.

Krutonium:
This is exactly what I need on Linux right now. Could you point me at a built version, or at an easy to follow way to build it myself?

PR #256, opened 2020-12-16
Initialize CEF early to fix macOS crash

eric:
### Description

Initialize CEF during `obs_module_load()` to prevent a crash in Chrome's memory allocation handler.

### Motivation and Context

If CEF is initialized after CoreAudio on macOS Big Sur, it trips a bug in Chrome that causes a crash.

I'm tired of my OBS crashing unless I specifically open a Scene Collection with only a Browser in it before opening the Scene Collections I want to open.

### How Has This Been Tested?

I opened OBS and it doesn't crash.

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-plugin.cpp
+++ b/obs-browser-plugin.cpp
@@ -592,6 +592,12 @@ bool obs_module_load(void)
 	}
 	obs_data_release(private_data);
 #endif
+
+#if defined(__APPLE__) && CHROME_VERSION_BUILD < 4183
+	// Make sure CEF malloc hijacking happens early in the process
+	obs_browser_initialize();
+#endif
+
 	return true;
 }
 


PatTheMav:
As we plan to fix this with an updated CEF version that I've been working on, could you add some conditionals for this to only take effect when CEF version is lower or equal to 3770? This would allow us to merge this as-is and not have any side-effects while we update the plugin for our new CEF version.

obs-browser-plugin.cpp
```diff
@@ -592,6 +592,12 @@ bool obs_module_load(void)
 	}
 	obs_data_release(private_data);
 #endif
+
+#if defined(__APPLE__)
```
WizardCM, line 596: I'd suggest

```cpp
#if defined(__APPLE__) && CHROME_VERSION_BUILD < 4183
```

That way this code is discarded when we upgrade CEF.

Ideally, I'd also prefer we had a runtime check for the macOS version so that this code only runs on Big Sur to avoid side effects on older versions.

eric: Updated.

eric:
I've updated it with the conditional @WizardCM suggested.

PatTheMav:
Sounds good enough to me - `obs_browser_initialize` will be called again in `RegisterBrowserSource` but the initialize method itself uses the `manager_initialized` to only run once, so there should be no harm.

daviscabral:
Thank you sir, I have a custom build running since this was merged and it is working pretty well.

franckhlmartin:
> Thank you sir, I have a custom build running since this was merged and it is working pretty well.

@daviscabral That's very good to know. Did you do the custom build or is it available somewhere? Can you please share?

Issue #265, opened 2021-02-11
Browse Panel and web page size inconsistent


mhartington:
## Platform

Operating system and version: macOS 11.2.1
OBS Studio version: 26.1.2-141-g31a9dc384 (built from source)

## Expected Behavior

A browser panel and the loaded webpage should be the same size once loaded or if the panel is resized

## Current Behavior

Building from source to test the new browser panel. If a web page is loaded the size of the embedded web page will be greater than the actual panel size from OBS. 

Here in this gif, I'm loading a Twitch dashboard, but the rendered window is much larger than the actual window UI 
https://user-images.githubusercontent.com/2835826/107712632-be176d80-6c97-11eb-8fb8-f2ca99031661.mov

## Steps to Reproduce
1. Open version of OBS on mac with browser panel dock enabled
2. create new panel pointing to a web page 
    I created this sample test that just prints the JS-window to the html https://cdpn.io/mhartington/debug/LYbbrpa/xnMabmZNEpLr
3. Open the dock and try to resize the panel
4. Open the URL in your browse and observe the size report differenec

## Additional information

Here are two additional screenshots from the url above, one is fullscreen on a 16" macbook pro and the other is from obs, the size difference is very different.

<img width="2048" alt="Screen Shot 2021-02-11 at 6 45 24 PM" src="https://user-images.githubusercontent.com/2835826/107713684-cd97b600-6c99-11eb-88e6-2881fd598533.png">

<img width="2048" alt="Screen Shot 2021-02-11 at 6 45 18 PM" src="https://user-images.githubusercontent.com/2835826/107713686-cd97b600-6c99-11eb-80b9-53ef721d9986.png">

Sorry in advanced if this is not the right place to post the issue. Let me know if I should report this to the main OBS repo instead.

Session Logfile

<details>

```
18:59:53.866: CPU Name: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
18:59:53.866: CPU Speed: 2300MHz
18:59:53.866: Physical Cores: 8, Logical Cores: 16
18:59:53.866: Physical Memory: 32768MB Total
18:59:53.866: OS Name: Mac OS X
18:59:53.866: OS Version: Version 11.2.1 (Build 20D74)
18:59:53.866: Kernel Version: 20.3.0
18:59:53.868: hotkeys-cocoa: Getting keyboard keys failed
18:59:53.868: hotkeys-cocoa: Getting keyboard keys failed
[...]

PR #267, opened 2021-02-21
Fix unnecessary pixel ratio calculations on macOS

PatTheMav:
### Description
Disables unnecessary pixel ratio calculations on macOS which led to CEF window size and Qt window size mismatches.

Fixes https://github.com/obsproject/obs-browser/issues/265.

### Motivation and Context
Window sizes in macOS are automatically translated to actual pixel sizes internally by macOS, applications are required to work with "logical" sizes (aka "dots") instead. Thus applying pixel ratios as multipliers is not necessary.

### How Has This Been Tested?
* Applied to local build on macOS and tested with attached and detached browser panels.

### Types of changes

### Checklist:

### DIFF
--- a/panel/browser-panel-internal.hpp
+++ b/panel/browser-panel-internal.hpp
@@ -55,8 +55,10 @@ public:
 	std::string script;
 	CefRefPtr<CefRequestContext> rqc;
 	QTimer timer;
+#ifndef __APPLE__
 	QPointer<QWindow> window;
 	QPointer<QWidget> container;
+#endif
 	bool allowAllPopups_ = false;
 
 	virtual void resizeEvent(QResizeEvent *event) override;
--- a/panel/browser-panel.cpp
+++ b/panel/browser-panel.cpp
@@ -177,8 +177,10 @@ QCefWidgetInternal::QCefWidgetInternal(QWidget *parent, const std::string &url_,
 
 	setFocusPolicy(Qt::ClickFocus);
 
+#ifndef __APPLE__
 	window = new QWindow();
 	window->setFlags(Qt::FramelessWindowHint);
+#endif
 }
 
 QCefWidgetInternal::~QCefWidgetInternal()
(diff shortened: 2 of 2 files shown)


WizardCM:
I'll be honest, I'm not a fan of how much code gets duplicated. The browser code is hard enough to navigate.

PatTheMav:
> I'll be honest, I'm not a fan of how much code gets duplicated. The browser code is hard enough to navigate.

I know, but the alternative would've been even less readable imo:

```
void QCefWidgetInternal::Init()
{
#ifdef __APPLE__
    WId handle = winId();
#else
    WId handle = window->winId();

    QSize size = this->size();
#ifdef SUPPORTS_FRACTIONAL_SCALING
    size *= devicePixelRatioF();
#else
    size *= devicePixelRatio();
#endif
#endif

    bool success = QueueCEFTask([this, handle]() {
		CefWindowInfo windowInfo;

		/* Make sure Init isn't called more than once. */
		if (cefBrowser)
			return;
#ifdef __APPLE__
		QSize size = this->size();
		windowInfo.SetAsChild((CefWindowHandle)handle, 0, 0,
				      size.width(), size.height());
#else
#ifdef _WIN32
        RECT rc = {0, 0, size.width(), size.height()};
#else
        CefRect rc = {0, 0, size.width(), size.height()};
#endif
        windowInfo.SetAsChild((CefWindowHandle)handle, rc);
#endif

		CefRefPtr<QCefBrowserClient> browserClient =
			new QCefBrowserClient(this, script, allowAllPopups_);

		CefBrowserSettings cefBrowserSettings;
		cefBrowser = CefBrowserHost::CreateBrowserSync(
			windowInfo, browserClient, url, cefBrowserSettings,
#if CHROME_VERSION_BUILD >= 3770
			CefRefPtr<CefDictionaryValue>(),
#endif
			rqc);
	});
    
    if (success) {
        timer.stop();
#ifndef __APPLE__
		if (!container) {
			container =
				QWidget::createWindowContainer(window, this);
			container->show();
		}

		Resize();
#endif
	}
}
```

If you'd prefer this style, I can adapt the PR as such.

jp9000:
I think the alternative is ok, it's not that much less readable. Matt probably does have a good point. If it's okay, could you use that instead?

Also, could you put the motivation and context into the commit message? It's good to have commit messages with details about the commit so it's easier to know the context of the commit if there's by some chance a bug or something in that code that we only discover later.

PatTheMav:
I just force-pushed another update that removed the duplication - funnily enough the diff looks "worse".

Also ran some tests if we could get away with less of those macOS-specific changes, but this (no sub-window, getting the size within the CEF task, disabling the resize function entirely) seems to be the safest way to get it to work on macOS. From what I could tell the container and sub-window were necessary for resizing support, which works out-of-the-box on macOS.

WizardCM:
Yeah, sometimes you have to sacrifice a good diff in favour of a more readable end result.

PatTheMav:
@WizardCM Looks like the Ubuntu-fix for clang-format hasn't been applied on this project yet?

WizardCM (approved):
Verified this does not break Windows or Linux panels. I think this is good to merge.

PR #269, opened 2021-03-11
Fix cmake option set

RytoEX:
### Description

Fix Shared Texture option in cmake-gui
Fix USE_QT_LOOP option in cmake-gui

When using `set` without specifying the CACHE argument, a normal/local variable is set to the specified value. This results in the GUI showing the CACHE variable value (OFF), but the rest of the CMake script checks the normal variable value first. Thus, the script works as if `SHARED_TEXTURE_SUPPORT_ENABLED` is set to ON, even though cmake-gui shows it set to OFF.

By default, cmake-gui on Windows would show `SHARED_TEXTURE_SUPPORT_ENABLED` as OFF, as specified by the initial `option` call.  The `option` call would set a CACHE variable in CMake.  The subsequent `set` call would then set a normal/local CMake variable to ON.  When CMake checks variables, it checks for a normal variable value first, and uses that value if found.  Because of this, the script would continue on as if `SHARED_TEXTURE_SUPPORT_ENABLED` was ON, even though cmake-gui showed it as OFF.

This PR fixes the behavior for both `SHARED_TEXTURE_SUPPORT_ENABLED` and `USE_QT_LOOP`.

Before:
image

After:
image

Alternatively, we could use a temporary normal variable to set the value and use a single option call.  I think it's a bit messier with just a couple boolean values, but I'm open to either solution.
```
set(SHARED_TEXTURE_SUPPORT_ENABLED OFF)
if(WIN32 OR (APPLE AND NOT BROWSER_LEGACY))
    set(SHARED_TEXTURE_SUPPORT_ENABLED ON)
endif()
option(SHARED_TEXTURE_SUPPORT_ENABLED "Enable shared texture support for the browser plugin" ${SHARED_TEXTURE_SUPPORT_ENABLED})
unset(SHARED_TEXTURE_SUPPORT_ENABLED)
```

References:
https://cmake.org/cmake/help/v3.19/command/option.html
https://cmake.org/cmake/help/v3.19/command/set.html
https://cmake.org/cmake/help/v3.19/command/unset.html
https://cmake.org/cmake/help/v3.19/variable/CACHE.html
https://cmake.org/cmake/help/v3.19/manual/cmake-language.7.html#variables
https://cmake.org/cmake/help/v3.19/manual/cmake-language.7.html#variable-references
https://cmake.org/cmake/help/v3.19/policy/CMP0077.html

### Motivation and Context

Having the GUI reflect values inconsistent from what the CMakeLists script evaluates is bad.

### How Has This Been Tested?

I compiled OBS with obs-browser on Windows 10 64-bit and tested browser source and browser panels.  Everything seemed to work fine.

### Types of changes

### Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,9 +17,10 @@ if(NOT CEF_FOUND)
 	return()
 endif()
 
-option(SHARED_TEXTURE_SUPPORT_ENABLED "Enable shared texture support for the browser plugin" OFF)
 if(WIN32 OR (APPLE AND NOT BROWSER_LEGACY))
-    set(SHARED_TEXTURE_SUPPORT_ENABLED ON)
+    option(SHARED_TEXTURE_SUPPORT_ENABLED "Enable shared texture support for the browser plugin" ON)
+else()
+    option(SHARED_TEXTURE_SUPPORT_ENABLED "Enable shared texture support for the browser plugin" OFF)
 endif()
 
 option(BROWSER_PANEL_SUPPORT_ENABLED "Enables Qt web browser panel support" ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,7 @@ option(BROWSER_PANEL_SUPPORT_ENABLED "Enables Qt web browser panel support" ON)
 if(NOT APPLE)
 	option(USE_QT_LOOP "Runs CEF on the main UI thread alongside Qt instead of in its own thread" OFF)
 else()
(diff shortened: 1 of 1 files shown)


WizardCM (approved):
I'm happy with this change as-is.

RytoEX:
@PatTheMav Any objections here?  I know it's a quick and dirty fix and that your CMake rewrite will supersede it, but this fixes the issue for now.

PatTheMav:
Yeah, it's what I did in my rewrite anyway.

PR #285, opened 2021-04-08
Change CMake behavior if CEF is not found

tytan652:
Continuation of this [obs-studio PR](https://github.com/obsproject/obs-studio/pull/4483)

### Description
CMake will now emit a fatal error if CEF is not found rather than "disabling" obs-browser.
Same if X11 is missing

### Motivation and Context
Continuation of this [obs-studio PR](https://github.com/obsproject/obs-studio/pull/4483)

Many Linux packagers doesn't seem to build OBS with obs-vst and obs-browser, or maybe they don't know about them.
So I create this PR, to make the packager really decide to include or not those two.

### How Has This Been Tested?

I replace the obs-studio's submodule with a symlink to my local repo without setting CEF_ROOT_DIR

### Types of changes
- CMake changes

### Checklist:


GeorgesStavracas (approved):
I'm in favor of this change :)

tytan652:
I found this
``` CMake
if(UNIX AND NOT APPLE)
	find_package(X11 REQUIRED)
	if(NOT X11_FOUND)
		message(STATUS "X11 Not found -- obs-browser plugin disabled.")
		return()
	endif()
	include_directories("${X11_INCLUDE_DIR}")
endif()
```
Should I modify it too ?

PR #294, opened 2021-05-22
Add Browser Source Refresh Hotkey

gxalpha:
Adds a hotkey to the browser source that refreshes its cache, the same
way the button in the settings does

### Description

Adds a Hotkey to the browser source that, when triggered, refreshes the browsers cache.

### Motivation and Context

Saw a Discord conversation asking for a feature like this.

### How Has This Been Tested?

Compiled with current master of obs-studio
Triggering the hotkey will refresh the page as intended.

macOS 11.2.3

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -72,6 +72,20 @@ void DispatchJSEvent(std::string eventName, std::string jsonString,
 BrowserSource::BrowserSource(obs_data_t *, obs_source_t *source_)
 	: source(source_)
 {
+
+	/* Register Refresh hotkey */
+	auto refreshFunction = [](void *data, obs_hotkey_id, obs_hotkey_t *,
+				  bool pressed) {
+		if (pressed) {
+			BrowserSource *bs = (BrowserSource *)data;
+			bs->Refresh();
+		}
+	};
+
+	obs_hotkey_register_source(source, "ObsBrowser.Refresh",
+				   obs_module_text("RefreshNoCache"),
+				   refreshFunction, (void *)this);
+
 	/* defer update */
 	obs_source_update(source, nullptr);
 


WizardCM:
I personally don't see the use case for a feature like this. A browser source should not need to be refreshed manually. Especially not frequently enough for a hotkey.

gxalpha:
I would disagree that there is no use case regarding this:

>A browser source should not need to be refreshed manually

True. The problem is, not all webpages are created equally well (from own experience,). Some pages were not desigened to be used as browser sources, others are just buggy - both issues can often times be addressed by refreshing.

>Especially not frequently enough for a hotkey

The reasoning for a hotkey in this case isn't the frequency, but the lack of good alternatives. Currently, there are 3 (4) ways to refresh a site that needs occasional refreshes:
- Refresh on scene switch (or whatever its called): This, while doing the job, will cause the loading process of the page to be visible on the output. This looks horrible, especially if for example custom CSS is applied, which seems to be the last step of the process
- Manually refresh while the scene is active: This causes the same problem of the loading being visible, just less frequently. In addition, you may need to first tab out of something like a game, resulting in additional work.
- Manually refresh via studio mode: Besides potentially having to unfocus the current application, this means having to go to studio mode, clicking on the scene with the source, selecting the source, then clicking "Refresh" in the toolbar and potentially exiting studio mode again. It's a lot of clicks that could be simplified.
- (Refreshing browser sources with a Hotkey via [Exeldros lua script](https://github.com/exeldro/obs-lua/blob/master/refresh-browsers.lua). This is the same behavior as this PR, just that it refreshes all browser sources at once or needs small modifications to work on single sources. Personally, this is how I deal with refreshing browsers at the moment, its the most convinient, and it is the workflow that would be made much easier by this PR)

That is to say, I do think a hotkey can be justified even if its not going to be used very frequently.

PR #302, opened 2021-07-12
Remove old CEF ifdefs

gxalpha:
### Description

Removes ifdefs that only benefit anyone using a CEF version older than 3770.
Since 3770 is the oldest version of CEF supported and shipped, their content was dead code.

This dead code got removed, none of the current functionality on any CEF >= 3770 should be effected.

### Motivation and Context

Stumbled across some of them while working on something else, and noticed they weren't needed.

### How Has This Been Tested?

Compiled on macOS.
This did require me to rebuild the cmake project and clear Xcode caches since Xcode has some scripts that assumed the removed header was still there, but that isn't project-related.

I unfortunately can't test Windows. 

### Types of changes

### Checklist:

### DIFF
--- a/browser-app.cpp
+++ b/browser-app.cpp
@@ -50,16 +50,9 @@ CefRefPtr<CefBrowserProcessHandler> BrowserApp::GetBrowserProcessHandler()
 
 void BrowserApp::OnRegisterCustomSchemes(CefRawPtr<CefSchemeRegistrar> registrar)
 {
-#if CHROME_VERSION_BUILD >= 3683
 	registrar->AddCustomScheme("http",
 				   CEF_SCHEME_OPTION_STANDARD |
 					   CEF_SCHEME_OPTION_CORS_ENABLED);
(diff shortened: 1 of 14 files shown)


WizardCM:
As it stands, I'd personally prefer to make this change once we've actually moved away from 3770. Essentially, internally "support" one version below what's currently included, just in case things go haywire and we need to revert the CEF version for some reason.

gxalpha:
Should I leave this PR open then and just wait?

WizardCM:
Yep, leave it open and we'll review it closer to v28.

pkviet:
Lgtm

WizardCM:
After merging #323 this now has merge conflicts. Once they're solved, I'll merge this.

WizardCM:
Ah, sorry I didn't catch this earlier - I would like to keep the `ENABLE_WASHIDDEN` option, as the goal (which I realise may not be realistic) is to eventually switch back to the previous behaviour.

PR #290, opened 2021-04-19
Update CMakeLists for modern builds

PatTheMav:
### Description
Updates CMake build script to coincide with changes in https://github.com/obsproject/obs-studio/pull/4560. Header definitions have been changed in source files appropriately.

### Motivation and Context
Updates CMake build scripts last updated 7 years ago and improves build situation on macOS.

### How Has This Been Tested?
* Full OBS builds on macOS, Linux, and Windows

### Types of changes
- Code cleanup (non-breaking change which makes code smaller or more readable)
- Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #310, opened 2021-08-03
obs-browser: Groundwork to enable Chrome 4430/arm64/Apple silicon

Developer-Ecosystem-Engineering:
- Update for custom obs texture sharing support used in CEF
Since these CEF patches are not available on any particularly OBS Project repository, instructions are provided here

```
PREFIX=/Users/youruser/Documents/cef-4430
mkdir -p $PREFIX
cd $PREFIX
mkdir cef
cd cef
mkdir automate
mkdir chromium_git
git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
export PATH=$PREFIX/cef/depot_tools/:$PATH
cd automate
curl -O https://bitbucket.org/chromiumembedded/cef/raw/master/tools/automate/automate-git.py
cd ../chromium_git
python ../automate/automate-git.py --download-dir=$PREFIX/cef/chromium_git --depot-tools-dir=$PREFIX/cef/depot_tools --no-distrib --no-build --arm64-build --branch=4430
cd cef 
patch -p1  < viz_osr_d3d_4430.patch
cd ../chromium/src/cef/ 
./cef_create_projects.sh
cd ..
mv out/Release_GN_x64 out/Release_GN_arm64
vim out/Release_GN_arm64/args.gn
**edit** target_cpu="arm64"
ninja -C out/Release_GN_arm64 cef
cd cef/tools 
./make_distrib.sh --ninja-build --minimal --arm64-build
cd ../binary_distrib 
```
This should produce "cef_binary_90.6.7+g19ba721+chromium-90.0.4430.212_macosarm64_minimal.zip" which will be used as a part of the OBS Studio build and will also need these OBS Browser changes.

### Description
When using CEF 4430 or later, we should use the texture sharing/gpu compositing feature that OBS Browser relies on, otherwise behaves like previously

### Motivation and Context
OBS Studio relies on a custom patch to CEF in order to perform certain features in OBS Browser. Until these patches are integrated into an Apple silicon variant of CEF, OBS Studio isn't at feature parity natively on Apple silicon.

### How Has This Been Tested?
Tested on M1 iMac and M1 MBP 
macOS 11.5.1

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-plugin.cpp
+++ b/obs-browser-plugin.cpp
@@ -69,6 +69,7 @@ static int adapterCount = 0;
 static std::wstring deviceId;
 
 bool hwaccel = false;
+bool gpuCompositing = true;
 
 /* ========================================================================= */
 
@@ -350,7 +351,7 @@ static void BrowserInit(void)
 	}
 #endif
 
-	app = new BrowserApp(tex_sharing_avail);
+	app = new BrowserApp(tex_sharing_avail, gpuCompositing);
 
 #ifdef _WIN32
 	CefExecuteProcess(args, app, nullptr);
@@ -693,7 +694,7 @@ bool obs_module_load(void)
 
 #ifdef SHARED_TEXTURE_SUPPORT_ENABLED
 	obs_data_t *private_data = obs_get_private_data();
-	hwaccel = obs_data_get_bool(private_data, "BrowserHWAccel");
+	gpuCompositing = hwaccel = obs_data_get_bool(private_data, "BrowserHWAccel");
 
 	if (hwaccel) {
 		check_hwaccel_support();
--- a/obs-browser-plugin.cpp
+++ b/obs-browser-plugin.cpp
@@ -299,7 +299,7 @@ static void BrowserInit(void)
 	prod_ver << std::to_string(obs_maj) << "." << std::to_string(obs_min)
 		 << "." << std::to_string(obs_pat);
 
-#if CHROME_VERSION_BUILD >= 4472
+#if CHROME_VERSION_BUILD >= 4430
 	CefString(&settings.user_agent_product) = prod_ver.str();
 #else
 	CefString(&settings.product_version) = prod_ver.str();
@@ -694,7 +694,8 @@ bool obs_module_load(void)
 
 #ifdef SHARED_TEXTURE_SUPPORT_ENABLED
 	obs_data_t *private_data = obs_get_private_data();
-	gpuCompositing = hwaccel = obs_data_get_bool(private_data, "BrowserHWAccel");
+	gpuCompositing = hwaccel =
+		obs_data_get_bool(private_data, "BrowserHWAccel");
 
 	if (hwaccel) {
 		check_hwaccel_support();
--- a/obs-browser-plugin.cpp
+++ b/obs-browser-plugin.cpp
@@ -694,8 +694,8 @@ bool obs_module_load(void)
 
 #ifdef SHARED_TEXTURE_SUPPORT_ENABLED
 	obs_data_t *private_data = obs_get_private_data();
-	gpuCompositing = hwaccel =
-		obs_data_get_bool(private_data, "BrowserHWAccel");
+	gpuCompositing = obs_data_get_bool(private_data, "BrowserHWAccel");
+	hwaccel = gpuCompositing;
 
 	if (hwaccel) {
 		check_hwaccel_support();
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -134,6 +134,13 @@ bool BrowserSource::CreateBrowser()
 		if (hwaccel) {
 			obs_enter_graphics();
 			tex_sharing_avail = gs_shared_texture_available();
+#if CHROME_VERSION_BUILD >= 4430 && defined(__APPLE__) && defined(__arm64__)
+			// CEF 4430 doesn't seem to work correctly when init'd with shared texture on arm64:
+			//  - we never see OnAcceleratedPaint called
+			//  - everything is black
+			// so we'll disable that for the time being
+			tex_sharing_avail = false;
+#endif
 			obs_leave_graphics();
 		}
 #else
@@ -177,10 +184,10 @@ bool BrowserSource::CreateBrowser()
 		cefBrowserSettings.default_font_size = 16;
 		cefBrowserSettings.default_fixed_font_size = 16;
 
-#if ENABLE_LOCAL_FILE_URL_SCHEME
+#if ENABLE_LOCAL_FILE_URL_SCHEME && CHROME_VERSION_BUILD < 4430
 		if (is_local) {
 			/* Disable web security for file:// URLs to allow
-			 * local content access to remote APIs */
+              * local content access to remote APIs */
 			cefBrowserSettings.web_security = STATE_DISABLED;
 		}
 #endif
@@ -551,9 +558,13 @@ void BrowserSource::Render()
 	if (texture) {
 #ifdef __APPLE__
 		bool useDefaultRect = hwaccel;
-		gs_effect_t *effect =
-			obs_get_base_effect((useDefaultRect) ? OBS_EFFECT_DEFAULT_RECT
-						      : OBS_EFFECT_DEFAULT);
+#if CHROME_VERSION_BUILD >= 4430 && defined(__APPLE__) && defined(__arm64__)
+		useDefaultRect = false;
+		flip = false;
+#endif
+		gs_effect_t *effect = obs_get_base_effect(
+			(useDefaultRect) ? OBS_EFFECT_DEFAULT_RECT
+					 : OBS_EFFECT_DEFAULT);
 #else
 		gs_effect_t *effect = obs_get_base_effect(OBS_EFFECT_DEFAULT);
 #endif
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -141,7 +141,7 @@ bool BrowserSource::CreateBrowser()
 #endif
 
 		CefRefPtr<BrowserClient> browserClient = new BrowserClient(
-			this, hwaccel && tex_sharing_avail, reroute_audio);
+			this, tex_sharing_avail, reroute_audio);
 
 		CefWindowInfo windowInfo;
 #if CHROME_VERSION_BUILD < 3071
@@ -152,7 +152,7 @@ bool BrowserSource::CreateBrowser()
 		windowInfo.windowless_rendering_enabled = true;
 
 #ifdef SHARED_TEXTURE_SUPPORT_ENABLED
-		windowInfo.shared_texture_enabled = hwaccel;
+		windowInfo.shared_texture_enabled = tex_sharing_avail;
 #endif
 
 		CefBrowserSettings cefBrowserSettings;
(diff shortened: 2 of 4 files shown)


WizardCM:
Please split the changes for newer CEF support to a separate commit from the hardware acceleration related changes.

Additionally, please be sure to run clang-format.

Developer-Ecosystem-Engineering:
> Please split the changes for newer CEF support to a separate commit from the hardware acceleration related changes.
> 

You need a CEF that supports Apple silicon and the custom OBS code for enabling shared texture support is about hardware acceleration. I'm not sure we follow what you want split?

WizardCM (changes requested):
> what lines are you wanting as their own PR?

Not as their own PR, but as a separate commit in this PR. The idea is that each commit should be a single "functional" change, and it should be possible to revert them independently.

I've added a couple comments on the areas that should be their own commit, likely simply titled "Enable basic CEF 4430 support", whereas everything else would be "Enable better texture support on CEF ARM versions" as while these are related to each other, they have a different purpose.

Also, separate note - has this been tested on Windows? If scoping this to ARM removes any effect to Windows, then ignore this question.

Either way, thank you for your efforts. :)

gnuherdnix:
hi @Developer-Ecosystem-Engineering I'm attempting to build your PR, but I'm unable to find the patch you shared in your instructions: `viz_osr_d3d_4430.patch`. Has it been uploaded as part of a PR across one of the various OBS repositories?

Developer-Ecosystem-Engineering:
> hi @Developer-Ecosystem-Engineering I'm attempting to build your PR, but I'm unable to find the patch you shared in your instructions: `viz_osr_d3d_4430.patch`. Has it been uploaded as part of a PR across one of the various OBS repositories?

Hi @gnuherdnix, that patch can be found here https://bitbucket.org/chromiumembedded/cef/pull-requests/285, thats not our work but appears to be some custom work to support texture sharing in OBS Studio to improve performance and reduce CPU overhead on x86. It doesn't appear to be technically necessary on Apple silicon, but we are still reviewing some tests and data based on feedback from the team. We will need to rework this PR based on some feedback, that should happen in a bit.

pabloko:
Hi @Developer-Ecosystem-Engineering, this patch is not a performance enhancement, what it really does, is to bring back the OnAcceleratedPaint from CefRenderHandler, wich allows to access a gpu texture instead having to map raw RGBA pixels in memory, but most importantly, allows that each frame can be requested, this is very important for video in order to prevent tearing. In windows it uses DirectX11 shared handles and IOSurface on apple.
Do somebody know if this patch is already on the main windows obs-browser dist?

RytoEX:
> 
> 
> Hi @Developer-Ecosystem-Engineering, this patch is not a performance enhancement, what it really does, is to bring back the OnAcceleratedPaint from CefRenderHandler, wich allows to access a gpu texture instead having to map raw RGBA pixels in memory, but most importantly, allows that each frame can be requested, this is very important for video in order to prevent tearing. In windows it uses DirectX11 shared handles and IOSurface on apple.
> Do somebody know if this patch is already on the main windows obs-browser dist?

@pabloko 
On Windows, we use a prebuilt patched version of CEF 3770 ([x64](https://cdn-fastly.obsproject.com/downloads/cef_binary_75.1.16+g16a67c4+chromium-75.0.3770.100_windows64_minimal.zip), [x86](https://cdn-fastly.obsproject.com/downloads/cef_binary_75.1.16+g16a67c4+chromium-75.0.3770.100_windows32_minimal.zip)).  On macOS, we use a prebuilt patched version of [CEF 4183](https://cdn-fastly.obsproject.com/downloads/cef_binary_4183_macosx64.tar.bz2).

For all intents and purposes, to the end user, enabling hardware accelerated browser sources is a performance enhancement because it allows them to have fluid 60 FPS complex browser-based sources and overlays.  We're not talking about increasing the efficiency of a loop or function when we use the phrase in this context.

gnuherdnix:
@RytoEX out of curiosity, could you point me to where the build infrastructure and patches are stored for the CEF prebuilts used for Mac & Windows? I haven't been able to find them and wanted to try building locally.

@Developer-Ecosystem-Engineering Thanks for the link. Regarding IOSurfaces, in the example at hand where we render onto the IOSurface backed GL texture from the CEF process, and then read from that same texture for rendering to the scene in OBS, are we managing to work with the same pointer to on-device GPU memory and avoid CPU<->GPU memory copies? 

My understanding is that IOSurface allows transparent access to the surface data and will copy whether needed from GPU or CPU memory, but I'm wondering if we effectively avoid CPU copies when we are only reading & writing to the GPU memory buffer across the two processes.

RytoEX:
> @RytoEX out of curiosity, could you point me to where the build infrastructure and patches are stored for the CEF prebuilts used for Mac & Windows? I haven't been able to find them and wanted to try building locally.

We don't have a build infrastructure for our CEF prebuilts.  They were generously built by contributors so that we could host them so that people don't have to rebuild them when building obs-studio or obs-browser.

I don't believe we have any custom CEF patches publicly posted.  As @Developer-Ecosystem-Engineering has said, the texture sharing patch we use is that unmerged CEF PR.  It's possible that we modified it further to get it to actually work, but I didn't work on it, so I can't speak to the details.

obs-browser-source.cpp
```diff
@@ -177,12 +184,12 @@ bool BrowserSource::CreateBrowser()
...
-			/* Disable web security for file:// URLs to allow
-			 * local content access to remote APIs */
-			cefBrowserSettings.web_security = STATE_DISABLED;
-		}
+#if ENABLE_LOCAL_FILE_URL_SCHEME && CHROME_VERSION_BUILD < 4430
+     if (is_local) {
+             /* Disable web security for file:// URLs to allow
+              * local content access to remote APIs */
+             cefBrowserSettings.web_security = STATE_DISABLED;
+     }
```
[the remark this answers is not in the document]
Developer-Ecosystem-Engineering: Done

browser-client.cpp
```diff
@@ -182,6 +182,7 @@ void BrowserClient::OnPaint(CefRefPtr<CefBrowser>, PaintElementType type,
 
 #ifdef SHARED_TEXTURE_SUPPORT_ENABLED
 	if (sharing_available) {
+		blog(LOG_INFO, "[obs-browser]: BrowserClient::OnPaint: called when sharing_available");
```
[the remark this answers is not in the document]
Developer-Ecosystem-Engineering: Removed

obs-browser-plugin.cpp
```diff
@@ -298,7 +299,7 @@ static void BrowserInit(void)
 	prod_ver << std::to_string(obs_maj) << "." << std::to_string(obs_min)
 		 << "." << std::to_string(obs_pat);
 
-#if CHROME_VERSION_BUILD >= 4472
+#if CHROME_VERSION_BUILD >= 4430
```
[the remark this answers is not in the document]
Developer-Ecosystem-Engineering: Done

Developer-Ecosystem-Engineering:
- clang-format re-run on all the changes
- CEF version and HW Accel changes are split into two distinct commits 
- Removed debug log

obs-browser-source.cpp
```diff
@@ -134,14 +134,21 @@ bool BrowserSource::CreateBrowser()
 		if (hwaccel) {
 			obs_enter_graphics();
 			tex_sharing_avail = gs_shared_texture_available();
+#if CHROME_VERSION_BUILD >= 4430 && defined(__APPLE__) && defined(__arm64__)
+			// CEF 4430 doesn't seem to work correctly when init'd with shared texture on arm64:
+			//  - we never see OnAcceleratedPaint called
+			//  - everything is black
+			// so we'll disable that for the time being
```
RytoEX, line 141: There was a comment on CEF PR 285 recently about OnPaint was being called instead of OnAcceleratedPaint and they had to disable the sandbox to get this to work.
https://bitbucket.org/chromiumembedded/cef/pull-requests/285/reimplement-shared-texture-support-for-viz#comment-243274341

RytoEX (changes requested):
It looks like the first two commits cancel each other out.  Please squash your commits so that only the relevant commits remain.

obs-browser-source.cpp
Developer-Ecosystem-Engineering: Interesting, nice find, will take a look!  Not sure if you want potential M1 users to take a look at Apple silicon OBS Browser with/without this change.  Won't be able to get to it today.

Developer-Ecosystem-Engineering:
> It looks like the first two commits cancel each other out. Please squash your commits so that only the relevant commits remain.

Apologies, should be resolved now, didn't mean to close the PR!

obs-browser-plugin.cpp
```diff
@@ -693,7 +694,7 @@ bool obs_module_load(void)
 
 #ifdef SHARED_TEXTURE_SUPPORT_ENABLED
 	obs_data_t *private_data = obs_get_private_data();
-	hwaccel = obs_data_get_bool(private_data, "BrowserHWAccel");
+	gpuCompositing = hwaccel = obs_data_get_bool(private_data, "BrowserHWAccel");
```
RytoEX: This goes against the style guide we use as a reference.  See the Linux coding style guide linked from https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst#coding-guidelines:

> Don't put multiple statements on a single line unless you have something to hide:
> Don't put multiple assignments on a single line either. Kernel coding style is super simple. Avoid tricky expressions.

RytoEX (changes requested):
[Use imperative mood](https://chris.beams.io/posts/git-commit/) for the commit message subjects.

For example:
> Add support for CEF 4430

I can't properly think of a way to rephrase the first commit message subject.

Wrap the commit message bodies at 72 characters.  End sentences in the commit message body with periods.

A clarifying point about this:
> On Apple silicon its not necessary to use the custom CEF patch which enables texture sharing on x86, to achieve 60 fps. Its wired up but disabled (enabling it has no effect on __arm64__ && __APPLE__)

Some users use many browser sources at once, sometimes running heavy workloads (bit cups used to be the common example).  Hardware acceleration and texture sharing reduce CPU load overall in addition to achieving 60 FPS, not just aiming to allow one browser source to reach 60 FPS.

Developer-Ecosystem-Engineering:
> Some users use many browser sources at once, sometimes running heavy workloads (bit cups used to be the common example). Hardware acceleration and texture sharing reduce CPU load overall in addition to achieving 60 FPS, not just aiming to allow one browser source to reach 60 FPS.

Hardware acceleration is still enabled. If you disable it in CEF, you'll definitely notice even on M1.

obs-browser-plugin.cpp
Developer-Ecosystem-Engineering: Fixed

obs-browser-source.cpp
```diff
@@ -134,14 +134,21 @@ bool BrowserSource::CreateBrowser()
...
+			tex_sharing_avail = false;
+#endif
```
RytoEX, line 143: Just noticed this was within `obs_enter_graphics()`.  Though it is just one extra assignment, we try to do as little possible inside the graphics context as possible.  This looks like it can safely be done after `obs_leave_graphics()`.  Or you could rearrange the ifdefs, but it's probably less complicated to just move this one.

RytoEX (changes requested):
Heads up, there are merge conflicts.

Am I understanding correctly that this PR's main change is changing the `disable-gpu-compositing` CEF/Chromium switch from relying on the value of the `enable-gpu` command line switch to using the value of the "Enable Browser Source Hardware Acceleration" option and preventing failures when using CEF >= 4430?

Developer-Ecosystem-Engineering:
I believe @PatTheMav integrated the spirit of these changes into his PR already and CEF is building 'elsewhere' to be included as a binary payload for arm64 variants of OBS. Would be good for him to confirm he's happy with it and it needs no further iteration.

PatTheMav:
Yeah we've mainly used the CEF build instructions. We still need to revisit the coupling of enabling GPU compositing in CEF and texture sharing (because even with the latter disabled, CEF seems to still do the former on my M1 Mac).

But I'm happy to solve/revisit this in the scope of another PR.

Developer-Ecosystem-Engineering:
Closing then since this was integrated via #290 

Lets see what the feedback is on the M1 OBS build w.r.t. CEF and what needs further work, if anything.

PR #312, opened 2021-08-28
Add scene browser controls

gxalpha:
### Description

Adds the scene controls which are already described in the permissions string ("Advanced access to OBS (Change scenes, Start/Stop replay buffer, etc.)").

To access the scenes available or the current scene, you need READ_ONLY, to change the scenes you need ADVANCED permission.

While I was at it, I also added functions to get and change transitions. These follow the same pattern.

### Motivation and Context

See above.

### How Has This Been Tested?

macOS 12, compiled and run

All the functions can be called by JS and do what they are supposed to.
No memory leaks found.

### Types of changes

### Checklist:


RytoEX:
Bumping this to OBS Studio 27.2 per discussion with @WizardCM to allow more time to review without holding back 27.1.

gxalpha:
Made the `setTransition` function cleaner, now depends on
- obsproject/obs-studio#5504

browser-client.cpp
```diff
@@ -168,7 +169,7 @@ bool BrowserClient::OnProcessMessageReceived(
 		CefProcessMessage::Create("executeCallback");
 
 	CefRefPtr<CefListValue> args = msg->GetArgumentList();
-	args->SetInt(0, message->GetArgumentList()->GetInt(0));
+	args->SetInt(0, arguments->GetInt(0));
```
jp9000, line 269: I think it would probably be more ideal to rename these two variables, `arguments` and `args`, to something like `input_args` and `execute_args`. Having two variables in the same function with nearly identical names but different purposes is confusing to the reader.

browser-client.cpp
```diff
@@ -118,19 +118,80 @@ bool BrowserClient::OnProcessMessageReceived(
...
+		} else if (name == "getScenes") {
+			struct obs_frontend_source_list list = {{{0}}};
+			obs_frontend_get_scenes(&list);
+			std::vector<char *> scenes_vector;
+			for (size_t i = 0; i < list.sources.num; i++) {
+				obs_source_t *source = list.sources.array[i];
+				scenes_vector.push_back(
+					(char *)obs_source_get_name(source));
+			}
+			json = Json::array{scenes_vector};
```
jp9000, line 192: I tested this just to make sure that this would function as it is written, and unfortunately it won't. Rather than an array of strings, it'll create an array with one item containing an array of strings. So basically it's currently creating something like this:
```json
[["string1", "string2", "string3"]]
```
which probably isn't what you wanted. I'm glad I checked this to verify.

Use
```cpp
json = scenes_vector
```
instead, which fixes it.

browser-client.cpp
```diff
@@ -148,16 +209,52 @@ bool BrowserClient::OnProcessMessageReceived(
...
+		} else if (name == "getTransitions") {
+			struct obs_frontend_source_list list = {{{0}}};
+			obs_frontend_get_transitions(&list);
+			std::vector<char *> transitions_vector;
+			for (size_t i = 0; i < list.sources.num; i++) {
+				obs_source_t *source = list.sources.array[i];
+				transitions_vector.push_back(
+					(char *)obs_source_get_name(source));
+			}
+			json = Json::array{transitions_vector};
```
jp9000, line 221: Same array issue.

browser-client.cpp
```diff
@@ -118,19 +118,80 @@ bool BrowserClient::OnProcessMessageReceived(
...
+				{"recordingPaused",
+				 obs_frontend_recording_paused()},
+				{"replaybuffer",
+				 obs_frontend_replay_buffer_active()},
+				{"virtualcam",
+				 obs_frontend_virtualcam_active()}};
+		} else if (name == "getScenes") {
+			struct obs_frontend_source_list list = {{{0}}};
+			obs_frontend_get_scenes(&list);
+			std::vector<char *> scenes_vector;
```
jp9000, line 186: There shouldn't be any need to use `std::vector<char *>`. Should be fine with `std::vector<const char *>`.

browser-client.cpp
```diff
@@ -148,16 +209,52 @@ bool BrowserClient::OnProcessMessageReceived(
...
-				{"recordingPaused",
-				 obs_frontend_recording_paused()},
-				{"replaybuffer",
-				 obs_frontend_replay_buffer_active()},
-				{"virtualcam",
-				 obs_frontend_virtualcam_active()}};
+		} else if (name == "getTransitions") {
+			struct obs_frontend_source_list list = {{{0}}};
+			obs_frontend_get_transitions(&list);
+			std::vector<char *> transitions_vector;
```
jp9000, line 215: Same thing about the vector typename.

browser-client.cpp
```diff
@@ -148,16 +209,52 @@ bool BrowserClient::OnProcessMessageReceived(
...
+		} else if (name == "getProfiles") {
+			char **profiles = obs_frontend_get_profiles();
+			char **profile = profiles;
+			std::vector<char *> profiles_vector;
+			while (profile && *profile) {
+				profiles_vector.push_back(*profile);
+				profile++;
+			}
+			bfree(profiles);
+			json = Json::array{profiles_vector};
```
jp9000, line 237: Same thing about the double array issue.

browser-client.cpp
```diff
@@ -148,16 +209,52 @@ bool BrowserClient::OnProcessMessageReceived(
...
+			char **collections =
+				obs_frontend_get_scene_collections();
+			char **collection = collections;
+			std::vector<char *> collections_vector;
+			while (collection && *collection) {
+				collections_vector.push_back(*collection);
+				collection++;
+			}
+			bfree(collections);
+			json = Json::array{collections_vector};
```
jp9000, line 252: Double array issue again.

NimitzDEV:
I have a different question, I was wondering are those APIs can be available to Custom Browser Docks, tested on OBS 27.1.3 doesn't seem to support it. 
If you could support it to be used in Custom Browser Docks, it can give developers the chance to make a custom look control panel.
Would love to hear some advice from you.

gxalpha:
I would have to defer to @WizardCM on that I think.
Also this should probably either be an issue or a post on the [ideas page](ideas.obsproject.com) as it's not directly related to the PR.

notr1ch:
I'm not sold on the permissions for some of these. Since browser sources default to read-only access, enumerating scenes, collections and profiles presents a real privacy risk to users. For example if someone streams with multiple profiles and one is an adult content site, a browser source page could surreptitiously record this.

I think the permissions need to be more granular so there is a separation between read-only fixed data and read-only user data.

WizardCM (changes requested):
Tested all APIs and I can confirm they work correctly with zero memory leaks.

I understand where @notr1ch is coming from, and I'd be OK with scrapping the ability to read / switch scene collections & profiles entirely, as I feel the scope of this API should be limited to the current scene collection & doing cool things with that rather than replacing obs-websocket.

To answer @NimitzDEV's question, the current API is built around the browser source specifically and was simply never implemented in the browser dock code. Doing so is quite advanced & would require its own permission UI, both of which are out of scope for this PR but definitely something I'd like to aim towards someday.

gxalpha:
Removed scene collection and profile related functions.

notr1ch:
I would consider scene names potentially sensitive user data too, but removing that might be a bit much. I'd still like to see an extra permission for reading "user data".

gxalpha:
This would make it `All`, `Advanced`, `Basic`, `Read (User)`, `Read (OBS)`, `None` (in that order), right?
Also, would we be ok with breaking all existing setups (which probably aren't that many as of yet, considering the capabilities in 27.1 are rather limited), which currently are set to `Read` and make them `Read (OBS)`?
Unfortunately, I don't think we can just rely on changing the default and not touching things the user set manually, since due to a bug introduced with the initial undo/redo ([fixed only recently](https://github.com/obsproject/obs-studio/pull/5296)), right now approximately all defaults are also saved in the scene collection.

WizardCM:
@gxalpha I think that's fine, let's do it.

gxalpha:
Added the new permission and tested it works.
image


I also *hope* the JSDoc definitions are correct, but I'm not sure (I don't know how to test them :D). Perhaps @VodBox wants to take a quick look.

PR #313, opened 2021-09-02
Automatically fetch CEF

mbrunnen:
When on Linux (`UNIX AND NOT APPLE`) and `CEF_ROOT_DIR` is not set, this change will try to download CEF with cmake's `FetchContent`.  I tried successfully a build on `Debian GNU/Linux 10 (buster)` with fetching CEF.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.12)
 project(obs-browser)
 
+include(FetchContent)
+
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}")
 set(CMAKE_INCLUDE_CURRENT_DIR TRUE)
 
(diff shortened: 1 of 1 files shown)


tytan652:
At least check for x86_64 architecture before trying to fetch

WizardCM:
I personally like this idea as a way to lower the barrier for building on Linux.

I do worry that it's yet another place to update dependency URLs when we update dependencies.

derrod:
This should probably not be enabled by default.

mbrunnen:
It is not enabled per default, only when CEF_ROOT_DIR is not set, while trying to build the browser function.

I do not see why fetching should only be done on x86_64.

Handling the URL (and hash) is a good point, where would these be handled?

mbrunnen:
> Please fill out the PR template.

Will do

jp9000:
Hi there, we'd prefer not to do this for now. But thank you very much in either case.

mbrunnen:
ok, your loss ;)

PR #325, opened 2021-10-27
Add lock for modify BrowserSource::cefBrowser

walker-WSH:
### Description

I meet a crash that caused by threadsafe of BrowserSource::cefBrowser.

_obs_graphics_thread_ is getting this shared-ptr variable in _BrowserSource::ExecuteOnBrowser_. 
While calling virtual _AddRef()_,  app crashed at __purecall()._

Check the CEF main thread, I find code location stay at _BrowserSource::CreateBrowser()_:
_cefBrowser = CefBrowserHost::CreateBrowserSync(...)_

Obviously, while changing cefBrowser, lock is need.

### Motivation and Context

make sure threadsafe

### How Has This Been Tested?

changing url/resolution frequently

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -116,7 +116,7 @@ void BrowserSource::ExecuteOnBrowser(BrowserFunc func, bool async)
 		}
 		os_event_destroy(finishedEvent);
 	} else {
-		CefRefPtr<CefBrowser> browser = cefBrowser;
+		CefRefPtr<CefBrowser> browser = GetBrowser();
 		if (!!browser) {
 #ifdef USE_QT_LOOP
 			QueueBrowserTask(cefBrowser, func);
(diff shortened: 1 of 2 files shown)


WizardCM:
Hi there, please rebase this PR.

Additionally, is this only an issue in browser sources & not browser docks/panels?

I wonder if a better solution would be to not destroy/recreate the entire source when changing URL or resizing, though having this solution as backup in other situations may be a good idea regardless.

walker-WSH:
> Hi there, please rebase this PR.
> 
> Additionally, is this only an issue in browser sources & not browser docks/panels?
> 
> I wonder if a better solution would be to not destroy/recreate the entire source when changing URL or resizing, though having this solution as backup in other situations may be a good idea regardless.

While cefBrowser in docks/panels is creating, if it may be accessed outsides CEF's main thread, I think it also needs a lock.
The way you said that not destroy/recreate the entire source can only reduce this issue.

PR #344, opened 2022-01-15
Fix sources not rendering after hide/show

### LINKED ISSUE #342: Browser source doesn't show on scene change
### Operating System Info

Ubuntu 20.04

### Other OS

_No response_

### OBS Studio Version

27.2.0 Beta 2

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/GWNAHFe1jwEmMJkH

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Browser source keep showing/working on scene change from a scene where that source wasn't present.

### Current Behavior

When having multiple scenes, where one scene has as source another scene, while the other scene contains a browser source, that browser source stops showing when changing to a scene where that scene with the browser source wasn't included.
When switching from the scene containing the browser source to a scene that is having the previous scene included, it keeps working.
Unfortunately there is nothing in the log file that indicates an error.
It seems not to happen, when I configure the browser source to unload when it's not visible.

### Steps to Reproduce

1. create a scene with a browser source
2. create another scene without that browser source
3. create a third scene including a source referencing the first scene (with the browser source, source NOT set to unload when invisible)
4. switch between scenes and observe how it stops being displayed when switching to the 2nd scene (that doesn't reference the browser source scene) and back
5. reloading the browser source (by making it invisible/visible again, or clearing cache) shows the source again

_No response_

WizardCM:
### Description

Back in 2019 when we updated to CEF 3770, the `WasHidden()` API [was broken](https://bitbucket.org/chromiumembedded/cef/issues/2483/osr-invalidate-does-not-generate-frame). 

Because of this, @zavitax had to write [a custom visibility handler](https://github.com/obsproject/obs-browser/pull/168). This worked well until the update to 4638, where it introduced terrible performance when sources weren't visible, so @VodBox [submitted a PR](https://github.com/obsproject/obs-browser/pull/330) to revert back to using the `WasHidden()` API, as it looked to be fixed.

Unfortunately, it seems the current CEF version has the [same bug as before](https://bitbucket.org/chromiumembedded/cef/issues/2483/osr-invalidate-does-not-generate-frame). Marshall has apparently [merged a fix](https://bitbucket.org/chromiumembedded/cef/pull-requests/409/), but the fix doesn't seem to work in this case.

**However**, the hack that previously didn't work - `HasResized()` - does seem to work now. This PR enables that hack.

In summary:

* pre-3770, we used `WasHidden()` with no issue
* for 3770, `WasHidden()` broke, so we used a custom visibility handler
* for 4638, the custom handler caused a performance regression so we re-enabled `WasHidden()`
* unfortunately, that introduced a bug on show, so now I've written this PR

------

~~Additionally, the PR that introduced 4638 support to obs-browser incorrectly renamed some variables. I need @VodBox and @PatTheMav to confirm whether this should be scoped to `_WIN32` or not.~~

### Motivation and Context

We like browser sources that function correctly & have good performance.

Fixes #342

### How Has This Been Tested?

* Run with or without hardware acceleration, hide source, wait, show source

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -52,6 +52,7 @@ static void SendBrowserVisibility(CefRefPtr<CefBrowser> browser, bool isVisible)
 
 #if ENABLE_WASHIDDEN
 	if (isVisible) {
+		browser->GetHost()->WasResized();
 		browser->GetHost()->WasHidden(false);
 		browser->GetHost()->Invalidate(PET_VIEW);
 	} else {


VodBox:
The flag should not be _WIN32. The reason the flag exists is to conditionally turn it on in CMake with a manual override. The reason it's not a convenient toggle is due to the fact that SendExternalBeginFrame at this time does not work.

WizardCM:
Alrighty, I'll ignore that part for now.

PR #345, opened 2022-01-15
Fix showing stale frames on show

VodBox:
### Description

This change aims to prevent stale frames being shown when a browser source is shown again after being hidden.

### Motivation and Context

Stale frames should not be shown.

### How Has This Been Tested?

Recording various browser pages being hidden then shown again, and checking that the shown frame doesn't match the frame from being hidden.

### Types of changes

### Checklist:


WizardCM:
Verified this fixes the issue. Very nice!

PR #364, opened 2022-05-30
obs-browser: Unsure saved source is alive

walker-WSH:
### Description

In browser-client, created source is saved into two list:
```
std::vector<obs_source_t*> audio_sources;
std::unordered_map<int, AudioStream> audio_streams;
```
But there is no reference for `audio_sources`.
I meet a crash as below:

### Checklist:

- [ ] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [ ] All commit messages are properly formatted and commits squashed where appropriate.


gxalpha:
Is there any particular reason why you're still running a CEF older than 4103?
I'm asking since we're looking to update our CEF again for the next version of OBS, in which case this code would get removed (we're usually supporting the current plus the last CEF).

walker-WSH:
> Is there any particular reason why you're still running a CEF older than 4103? I'm asking since we're looking to update our CEF again for the next version of OBS, in which case this code would get removed (we're usually supporting the current plus the last CEF).

since the crash in #360  ， I am still using the old CEF

RytoEX:
Closing per above. This code no longer exists and we no longer support CEF verions older than 5060.

PR #368, opened 2022-07-18
Fix macro-redefine warning with Qt loop enabled

tytan652:
### Description

Undefine libobs `UNUSED_PARAMETER` when Qt loop is enabled.

### Motivation and Context

I'm actually working on a PR to treat warnings as errors.

### How Has This Been Tested?

The warning/error no longer appear while building OBS Studio with this PR.

### Types of changes

### Checklist:


gxalpha:
To me it feels like the better fix would be using the existing macro instead of redefining a new one.

tytan652:
The two macros are different.

tytan652:
Waiting for obs-studio CI to check if the commit works.

PR #374, opened 2022-08-05
Null check macOS window handle before detaching view

gxalpha:
### Description

Fixes a crash with service integrations.
What the original code does is decoupling the browser from the widget (more explanation is in the comment a few lines above the modified). For some reason however, on Qt 6 the window handle can be null sometimes. In that case, trying to detach the view from the view from the superview fails (obviously).

By the way, the code looks cursed for two reasons:
1) It's Objective-C in a C(++) file
2) It was unreadable even before. If desired I can untangle the original code.

### Motivation and Context

Don't want crashes.

### How Has This Been Tested?

macOS 13.
No more crashes with Twitch integration.

### Types of changes

### Checklist:

### DIFF
--- a/panel/browser-panel.cpp
+++ b/panel/browser-panel.cpp
@@ -197,9 +197,10 @@ void QCefWidgetInternal::closeBrowser()
 		}
 #elif __APPLE__
 		// felt hacky, might delete later
-		id view = (id)cefBrowser->GetHost()->GetWindowHandle();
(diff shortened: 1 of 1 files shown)


WizardCM (approved):
Conceptually this matches the `if (hwnd) {` on Windows, so it looks good to me.

WizardCM:
Also we can save untangling the code for post-28 in my opinion.

PR #387, opened 2023-01-03
Add proc handler for javascript event

exeldro:
### Description
Add proc handler for javascript event to a browser source

### Motivation and Context
Allow other plugins to use javascript events on browser sources.
This can be used by the Browser Transition plugin to send start transition event including duration

### How Has This Been Tested?
With test code in the Browser Transition plugin

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -87,6 +87,22 @@ BrowserSource::BrowserSource(obs_data_t *, obs_source_t *source_)
 				   obs_module_text("RefreshNoCache"),
 				   refreshFunction, (void *)this);
 
+	auto jsEventFunction = [](void *p, calldata_t *calldata) {
+		const auto eventName = calldata_string(calldata, "eventName");
+		if (!eventName)
+			return;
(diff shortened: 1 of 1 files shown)


obs-browser-source.cpp
```diff
@@ -87,6 +87,18 @@ BrowserSource::BrowserSource(obs_data_t *, obs_source_t *source_)
 				   obs_module_text("RefreshNoCache"),
 				   refreshFunction, (void *)this);
 
+	auto jsEventFunction = [](void *p, calldata_t *calldata) {
+		DispatchJSEvent(calldata_string(calldata, "eventName"),
```
tt2468, line 91: `calldata_string()` can return NULL, and `DispatchJSEvent` takes an `std::string`, so it would be nice to have some basic validation for these fields. Consider a situation where no JSON string is defined, if an event does not require data.

exeldro: Added NULL checks for both eventName and jsonString

PR #391, opened 2023-02-04
Fix issues with failed JSON parsing in JS API responses

### LINKED ISSUE #390: Special characters seem to affect getScenes and getTransitions
### Operating System Info

Windows 11

### Other OS

_No response_

### OBS Studio Version

29.0.1

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/BdNarNPBbQaO1YOz

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Whenever using `window.obsstudio.getScenes` or `window.obsstudio.getTransitions` in OBS, I expect to receive the list of scenes or transitions available.

### Current Behavior

If one uses `window.obsstudio.getScenes` or `window.obsstudio.getTransitions` whenever a scene with special characters such as ', ", \ the callback passed to the functions is not invoked.

### Steps to Reproduce

1. Start OBS with remote debugging `$ ./obs64.exe --remote-debugging-port=9222`
2. In any scene, create a browser source with Page permissions 'Read access to user information'
3. Check for the list of scenes from the inspector (`http://localhost:9222/`) by running `window.obsstudio.getScenes((a) => {console.log(a)})` for the previously created browser source. The callback will be invoked `> ['OverlayTest', 'Webcam', 'Desktop', 'Demo', 'Other']`
4. Create a scene with one of these characters: ' , ", \
5. Perform step 3 again, this time the callback will not be invoked.

The same happens for transitions `getTransitions`

_No response_

WizardCM:
### Description

Using a [forum post](https://magpcss.org/ceforum/viewtopic.php?f=10&t=14807#p33456) as reference, overhauled JSON handling internally to use CefValue rather than the V8 context.

Note: It seems the V8 context is still required in this code block. I am unsure exactly why (it crashes otherwise).

### Motivation and Context

When users call API functions, they expect a response. Especially a valid one. Fixes #390.

### How Has This Been Tested?

`obsstudio.getScenes` and `obsstudio.getCurrentScene` with scene names that contain characters such as `"` and `\`.

Also re-tested all other get functions to make sure nothing else was broken.

### Types of changes

### Checklist:

### DIFF
--- a/browser-client.cpp
+++ b/browser-client.cpp
@@ -204,7 +204,7 @@ bool BrowserClient::OnProcessMessageReceived(
 		if (name == "getScenes") {
 			struct obs_frontend_source_list list = {};
 			obs_frontend_get_scenes(&list);
-			std::vector<const char *> scenes_vector;
+			std::vector<nlohmann::json> scenes_vector;
 			for (size_t i = 0; i < list.sources.num; i++) {
 				obs_source_t *source = list.sources.array[i];
 				scenes_vector.push_back(
@@ -230,7 +230,7 @@ bool BrowserClient::OnProcessMessageReceived(
 		} else if (name == "getTransitions") {
 			struct obs_frontend_source_list list = {};
 			obs_frontend_get_transitions(&list);
-			std::vector<const char *> transitions_vector;
+			std::vector<nlohmann::json> transitions_vector;
 			for (size_t i = 0; i < list.sources.num; i++) {
 				obs_source_t *source = list.sources.array[i];
 				transitions_vector.push_back(
--- a/browser-app.cpp
+++ b/browser-app.cpp
@@ -243,6 +243,56 @@ void BrowserApp::SetDocumentVisibility(CefRefPtr<CefBrowser> browser,
 }
 #endif
 
+CefRefPtr<CefV8Value> CefValueToCefV8Value(CefRefPtr<CefValue> value)
+{
+	CefRefPtr<CefV8Value> result;
+	switch (value->GetType()) {
+	case VTYPE_INVALID:
+		result = CefV8Value::CreateNull();
+		break;
+	case VTYPE_NULL:
+		result = CefV8Value::CreateNull();
+		break;
+	case VTYPE_BOOL:
+		result = CefV8Value::CreateBool(value->GetBool());
+		break;
+	case VTYPE_INT:
+		result = CefV8Value::CreateInt(value->GetInt());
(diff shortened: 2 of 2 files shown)


browser-client.cpp
```diff
@@ -206,13 +206,13 @@ bool BrowserClient::OnProcessMessageReceived(
...
 			obs_frontend_get_scenes(&list);
-			std::vector<const char *> scenes_vector;
+			std::vector<Json> scenes_vector;
 			for (size_t i = 0; i < list.sources.num; i++) {
 				obs_source_t *source = list.sources.array[i];
 				scenes_vector.push_back(
 					obs_source_get_name(source));
 			}
-			json = scenes_vector;
+			json = Json::array(scenes_vector);
```
tytan652, line 215: nlohmann is smart enough to convert a vector of JSONs to a JSON array of objects. And `Json::array` is json11 only.
```suggestion
			json = scenes_vector;
```

browser-client.cpp
```diff
@@ -234,13 +234,13 @@ bool BrowserClient::OnProcessMessageReceived(
...
 			obs_frontend_get_transitions(&list);
-			std::vector<const char *> transitions_vector;
+			std::vector<Json> transitions_vector;
 			for (size_t i = 0; i < list.sources.num; i++) {
 				obs_source_t *source = list.sources.array[i];
 				transitions_vector.push_back(
 					obs_source_get_name(source));
 			}
-			json = transitions_vector;
+			json = Json::array(transitions_vector);
```
tytan652, line 243: nlohmann is smart enough to convert a vector of JSONs to a JSON array of objects. And `Json::array` is json11 only.
```suggestion
			json = transitions_vector;
```

tytan652 (commented):
You might need to check if conversion are done correctly. But it should be good once the added json11 stuff is removed.

browser-client.cpp
WizardCM: Confirmed. Scrapped the change on my side.

browser-client.cpp
WizardCM: Confirmed. Scrapped the change on my side.

WizardCM:
Rebased on master (after fixing Nlohmann JSON compilation on Windows & macOS), removed unnecessary changes based on @tytan652's feedback, and retested with Nlohmann JSON specifically. It has the same output difference as json11 so this PR is still necessary to fix the linked issue.

Ready to be reviewed.

PR #394, opened 2023-02-10
cmake: Update CMake files for build system 3.0 update

PatTheMav:
### Description
This PR updates the CMake build files to be compatible with the build system 3.0 update on the main repository.

### Motivation and Context

* The current build script is retained as the "legacy" script (which will still be used to generate build projects on Linux and macOS)
* The new build script is used for macOS builds by default at first

The updated build system enforces uses Xcode by default on macOS and has some stricter code requirements, which are also fixed by this PR:

* Shadowing is fixed (new variables created in local scope with the same name as existing instance variables, thus shadowing them).
* Implicit casts are changed to be explicit (mainly for data being passed from data types with different bit-depth) - while data loss is still possible, the casts make it explicit that the developer is aware of this possible issues and wants to ignore it.

More information shall be available in the pull request on the main repository.

Note: This PR needs to be merged first, then the associated commit hash updated for the obs-browser submodule at obs-studio.

### How Has This Been Tested?
* New build scripts were tested as part of the overall update on macOS 13, Ubuntu 22.10, and Windows 11.

### Types of changes

### Checklist:
-  [x] I have read the [Contributing Guidelines](https://github.com/obsproject/obs-websocket/wiki/Contributing-Guidelines).
-  [x] All commit messages are properly formatted and commits squashed where appropriate.
-  [x] My code is not on `master` or a `release/*` branch.
-  [x] The code has been tested.
-  [x] I have included updates to all appropriate documentation.

### DIFF
--- a/.cmake-format.json
+++ b/.cmake-format.json
@@ -1,13 +1,46 @@
 {
-    "additional_commands": {
-      "find_qt": {
-        "flags": [],
-        "kwargs": {
-          "COMPONENTS": "+",
-          "COMPONENTS_WIN": "+",
-          "COMPONENTS_MACOS": "+",
-          "COMPONENTS_LINUX": "+"
+  "format": {
+    "line_width": 120,
+    "tab_size": 2,
+    "enable_sort": true,
+    "autosort": true
+  },
+  "additional_commands": {
+    "find_qt": {
+      "flags": [],
+      "kwargs": {
+        "COMPONENTS": "+",
+        "COMPONENTS_WIN": "+",
+        "COMPONENTS_MACOS": "+",
+        "COMPONENTS_LINUX": "+"
+      }
+    },
+    "set_target_properties_obs": {
+      "pargs": 1,
+      "flags": [],
+      "kwargs": {
+        "PROPERTIES": {
+          "kwargs": {
+            "PREFIX": 1,
+            "OUTPUT_NAME": 1,
+            "FOLDER": 1,
+            "VERSION": 1,
+            "SOVERSION": 1,
+            "FRAMEWORK": 1,
+            "BUNDLE": 1,
+            "AUTOMOC": 1,
+            "AUTOUIC": 1,
+            "AUTORCC": 1,
+            "AUTOUIC_SEARCH_PATHS": 1,
+            "BUILD_RPATH": 1,
+            "INSTALL_RPATH": 1,
+            "XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC": 1,
+            "XCODE_ATTRIBUTE_CLANG_WARN_SUSPICIOUS_IMPLICIT_CONVERSION": 1,
+            "XCODE_ATTRIBUTE_GCC_WARN_SHADOW":1 ,
+            "LIBRARY_OUTPUT_DIRECTORY": 1
+          }
         }
       }
     }
+  }
 }
--- a/cmake/os-windows.cmake
+++ b/cmake/os-windows.cmake
@@ -1,0 +1,34 @@
+target_compile_options(obs-browser PRIVATE $<IF:$<CONFIG:DEBUG>,/MTd,/MT>)
+target_compile_definitions(obs-browser PRIVATE ENABLE_BROWSER_SHARED_TEXTURE)
+
+target_link_libraries(obs-browser PRIVATE CEF::Wrapper CEF::Library d3d11 dxgi)
+target_link_options(obs-browser PRIVATE /IGNORE:4099)
+
+add_executable(obs-browser-helper WIN32 EXCLUDE_FROM_ALL)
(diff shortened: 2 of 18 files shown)


.cmake-format.json
```diff
@@ -1,13 +1,46 @@
...
-    "additional_commands": {
-      "find_qt": {
-        "flags": [],
-        "kwargs": {
-          "COMPONENTS": "+",
-          "COMPONENTS_WIN": "+",
-          "COMPONENTS_MACOS": "+",
-          "COMPONENTS_LINUX": "+"
+  "format": {
+    "line_width": 100,
```
RytoEX, line 3: 120

panel/browser-panel-client.cpp
```diff
@@ -76,11 +76,12 @@ void QCefBrowserClient::OnTitleChange(CefRefPtr<CefBrowser> browser,
 		if (title.compare("DevTools") == 0)
 			return;
 
+#if defined(_WIN32)
 		CefWindowHandle handl = browser->GetHost()->GetWindowHandle();
-#ifdef _WIN32
 		std::wstring str_title = title;
 		SetWindowTextW((HWND)handl, str_title.c_str());
 #elif defined(__linux__)
+		CefWindowHandle handl = browser->GetHost()->GetWindowHandle();
```
RytoEX, line 84: What is this fixing? Can it be split into its own PR?

RytoEX (changes requested):

cmake/windows/obs-browser-page.manifest
```diff
@@ -0,0 +1,27 @@
...
+                <requestedExecutionLevel
+                    level="asInvoker"
+                    uiAccess="false"
+                />
+            </requestedPrivileges>
+        </security>
+    </trustInfo>
+    <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+        <application>
+            <maxversiontested Id="10.0.19041.0"/>
```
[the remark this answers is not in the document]
PatTheMav: I've just copied the file as it currently exists in the repository - I've updated the values in the PR, but it seems the file was "unmaintained" so far anyway.

panel/browser-panel-client.cpp
PatTheMav: It fixes an "unused variable" error, because `handl` is only used on Windows and "shadowed" by the second assignment on Linux (so two violations in one).

PatTheMav: Given the small scope of the PR here I found it more convenient to bundle them here.

cmake/windows/obs-browser-page.manifest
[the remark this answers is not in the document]
RytoEX: Fair enough. I'm fine either way.

panel/browser-panel-client.cpp
RytoEX: Only used on Windows and Linux, leaving macOS the odd one out with the error, I think?

RytoEX: Either way, the change itself seems fine. I just like to try to separate the code changes from the CMake changes as much as possible to minimize review burdens and scope.

cmake/windows/obs-browser-page.manifest
```diff
@@ -0,0 +1,21 @@
...
+                <requestedExecutionLevel
+                    level="asInvoker"
+                    uiAccess="false"
+                />
+            </requestedPrivileges>
+        </security>
+    </trustInfo>
+    <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+        <application>
+            <maxversiontested Id="10.0.22000.0"/>
```
RytoEX, line 16: We can probably bump this up to whatever the latest dev system is on, but I don't know if it has any real effect.

RytoEX (commented):
One comment, but everything else looks fine.

RytoEX:
The commit messages may need to be adjusted. I don't see any changes to variable casting here. I think there were changes made to the scope after the initial draft.

cmake/windows/obs-browser-page.manifest
PatTheMav: Me neither, would require input on that.

cmake/os-windows.cmake
```diff
@@ -0,0 +1,34 @@
+target_compile_options(obs-browser PRIVATE $<IF:$<CONFIG:DEBUG>,/MTd,/MT>)
+target_compile_definitions(obs-browser PRIVATE ENABLE_BROWSER_SHARED_TEXTURE)
+
+target_link_libraries(obs-browser PRIVATE CEF::Wrapper CEF::Library d3d11 dxgi)
+target_link_options(obs-browser PRIVATE LINKER:/IGNORE:4099)
```
RytoEX, line 5: Is `LINKER:/IGNORE:4099` correct here?

PR #395, opened 2023-02-12
Increased maximum width and height

fungiboletus:
### Description
The maximum width and height are currently 4096 pixels, and this pull request double the maximum to 8192 pixels.

### Motivation and Context

This change is required to support higher resolutions, such as 7680x4320 (8K).

### How Has This Been Tested?
I compiled and run OBS on my laptop. I tested various webpages at low and high resolutions, including some webgl demos.

### Types of changes

### Checklist:

Feel free to close or reject my pull-request.

### DIFF
--- a/obs-browser-plugin.cpp
+++ b/obs-browser-plugin.cpp
@@ -186,9 +186,9 @@ static obs_properties_t *browser_source_get_properties(void *data)
 				OBS_TEXT_DEFAULT);
 
 	obs_properties_add_int(props, "width", obs_module_text("Width"), 1,
-			       4096, 1);
+			       8192, 1);
 	obs_properties_add_int(props, "height", obs_module_text("Height"), 1,
-			       4096, 1);
+			       8192, 1);
 
 	obs_properties_add_bool(props, "reroute_audio",
 				obs_module_text("RerouteAudio"));


WizardCM:
I have previously been hesitant about merging this change in particular as the supported value depends on the user's hardware and our graphics context (eg. D3D feature level).

As it stands, as we no longer support Windows 7 or 8, I'm more comfortable with this change.

RytoEX:
> As it stands, as we no longer support Windows 7 or 8, I'm more comfortable with this change.

Is that because browser hardware acceleration did not work on Windows 7, or some other reason? I wouldn't mind someone testing this on macOS and Linux, but I'm not opposed to this change in general. If macOS and Linux contributors think it's safe enough to land first and get wider testing later, I'm also not opposed.

Just for reference, the support for 8K textures depends on the DirectX hardware support or OpenGL implementation. In D3D 10.1 hardware (which OBS supposedly still supports, though [Windows 10 only requires DX9 hardware](https://support.microsoft.com/en-us/windows/windows-10-system-requirements-6d4e9a79-66bf-7950-467c-795cf0386715), while Windows 11 requires DX12 hardware), the maximum texture dimension is 8192, so this would be fine. In OpenGL, the maximum texture dimension is left to the implementation. Though GL_MAX_TEXTURE_SIZE below 8192 is exceedingly rare, they do still exist. For example, see the difference between these reports from August 2015 and August 2022, where the incidence of GL_MAX_TEXTURE_SIZE of below 8192 dropped from about 11% to below 1%:
* https://feedback.wildfiregames.com/report/opengl/feature/GL_MAX_TEXTURE_SIZE
* https://feedback.wildfiregames.com/gl/constants/#gl_max_texture_size

fungiboletus:
Perhaps we could fetch the value of `GL_MAX_TEXTURE_SIZE` once when loading the user interface and then use the lowest between it and 8192.

I will try to test on Mac OS, and perhaps some Linux distribution if I have enough time.

fungiboletus:
image

I tested on macOs 13.2.1 (arm64) and it works well.

derrod:
Basically all hardware still supported by OBS should support at least 16K x 16K texture sizes. So this PR seems fine to me.

RytoEX:
> Basically all hardware still supported by OBS should support at least 16K x 16K texture sizes. So this PR seems fine to me.

Debatable, as I've pointed out before that DX 10.1 supports only 8K, and we _supposedly_ still support DX 10.1 hardware on Windows 10. That being the case, bumping to 8K, as this PR does, is fine.

PR #401, opened 2023-03-23
Display a custom page for Dev Tools Discovery

WizardCM:
### Description

This enables a custom resource handler for CEF to allow overriding of individual internal files that CEF attempts to load.

The Dev Tools Discovery page is exposed via Remote Debugging and provides a list of all loaded browser pages to easily open Dev Tools from a single location.

This updated version introduces a number of enhancements:
1. Items are now ordered by category
2. Sub-pages are listed under parent pages
3. Built-in refresh button
4. OBS version number in page body based on User Agent
5. Clicking items opens them in a new tab by default
6. Custom styling

**Before:**

image

**After:**

image

### Motivation and Context

This file is normally packaged within CEF builds themselves. It's unnecessary to rebuild CEF just to change one HTML file.

Remote debugging is used for development and troubleshooting of browser sources.

### How Has This Been Tested?

Launch OBS with `--remote-debugging-port=1234` and visit the page.

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-plugin.cpp
+++ b/obs-browser-plugin.cpp
@@ -388,6 +388,10 @@ static void BrowserInit(void)
 
 	app = new BrowserApp(tex_sharing_avail);
 
+	BPtr<char> pagesPath = obs_module_file("dev_tools_discovery.html");
+	BPtr<char> pagesFile = os_quick_read_utf8_file(pagesPath);
+	app->SetDevToolsFile(pagesFile);
+
 #ifdef _WIN32
 	CefExecuteProcess(args, app, nullptr);
 #endif
--- a/browser-app.cpp
+++ b/browser-app.cpp
@@ -50,6 +50,17 @@ CefRefPtr<CefBrowserProcessHandler> BrowserApp::GetBrowserProcessHandler()
 	return this;
 }
 
+CefRefPtr<CefResourceBundleHandler> BrowserApp::GetResourceBundleHandler()
+{
+	return this;
+}
+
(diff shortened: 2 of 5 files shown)


obs-browser-plugin.cpp
```diff
@@ -388,6 +388,12 @@ static void BrowserInit(void)
 
 	app = new BrowserApp(tex_sharing_avail);
 
+	char *pagesPath = obs_module_file("dev_tools_discovery.html");
+	char *pagesFile = os_quick_read_utf8_file(pagesPath);
```
gxalpha, line 392: Could these be BPtr's instead (making the explicit bfree calls unneeded)?

WizardCM: Ah, I always forget about BPtr. Yes, that works here.

norihiro:
The two lines of the commit description have 73 characters.
```
The Dev Tools Discovery page is exposed via Remote Debugging and provides
a list of all loaded browser pages to easily open Dev Tools from a single
```

WizardCM:
Fixed commit description, and based on Warchamp's feedback off-thread I've removed the collapse capability as it's unnecessary.

Should be good to merge.

Lain-B:
Going to revert this for now based upon more feedback from Warchamp. You might have to remake this Matt. Sorry for the trouble. Also I should have spotted that you were using a raw pointer -- definitely want to use RAII and not manually free memory in the future. An std::string would have been sufficient for the `devToolsFile` variable.

PR #402, opened 2023-04-04
CI: Fix build errors with Xcode 14.3 and platform SDK 13.3

PatTheMav:
### Description
Makes warning about unqualified `std::` calls a non-error.

### Motivation and Context
Xcode 14.3 and the macOS 13.3 platform SDK introduced a few breaking changes:

* Updated AppleClang emits warnings about unqualified std cast calls when using C++ - as `move` is too broad a word, developers are to use `std::move` to make this explicit. Alas this is exactly what `json11` uses and because that library is archived, there is no possibility of an upstream update.

This will also affect any platform that uses recent `clang` to compile obs-studio, but for now it mainly seems to affect macOS.

### How Has This Been Tested?
Tested with Xcode 14.3 and AppleClang 14.0.3

### Types of changes

### Checklist:

### DIFF
--- a/cmake/os-macos.cmake
+++ b/cmake/os-macos.cmake
@@ -5,6 +5,9 @@ find_library(APPKIT AppKit)
 mark_as_advanced(COREFOUNDATION APPKIT)
 
 target_compile_definitions(obs-browser PRIVATE ENABLE_BROWSER_SHARED_TEXTURE ENABLE_BROWSER_QT_LOOP)
+if(CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 14.0.3)
+  target_compile_options(obs-browser PRIVATE -Wno-error=unqualified-std-cast-call)
+endif()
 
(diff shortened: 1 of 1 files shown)


cmake/os-macos.cmake
```diff
@@ -41,7 +42,7 @@ foreach(helper IN LISTS helper_suffixes)
     ${target_name}
     PROPERTIES MACOSX_BUNDLE_INFO_PLIST "${CMAKE_CURRENT_BINARY_DIR}/Info-Helper${helper_plist}.plist"
                OUTPUT_NAME "${target_output_name}"
-               FOLDER plugins/obs-browser/helper_suffixes
+               FOLDER plugins/obs-browser/Helpers
```
derrod, line 45: Is this an unrelated fix or was that also required (not explained in the PR description)?

RytoEX (approved):
Seems fine.

PR #408, opened 2023-05-03
Move browser-panel ObjC code to Objective-C file; Add more null checks

gxalpha:
### Description

Moves the Objective-C calls in `browser-panel.cpp` into an actual Objective-C file, making it more readable.
Also adds a `!view` null check in addition to the existing `!*((bool *)view)` which might fix a newly reported crash.

### Motivation and Context

https://github.com/obsproject/obs-studio/issues/8836 was reported and maybe the first null check fixes it. I have no idea if it actually does since I can't reproduce the crash in the first place.

### How Has This Been Tested?

macOS 13.3.1
Made sure that the crash of the original null check added in #374 didn't regress by using twitch integration with and without the call to `RemoveNSViewFromSuperview`.
Whether or not the new crash is fixed I don't know, I guess we'll see when we deploy to production.

### Types of changes

### Checklist:

### DIFF
--- a/panel/browser-panel-macos.m
+++ b/panel/browser-panel-macos.m
@@ -1,0 +1,15 @@
+#import <AppKit/AppKit.h>
+
+void RemoveNSViewFromSuperview(void *view)
+{
+	if (!view)
+		return;
+
+	if (!*((bool *)view))
+		return;
+
+	// CefBrowserHostView is a subclass of NSView
+	// and we just need the superclass here
+	NSView *nsview = (NSView *)view;
+	[nsview removeFromSuperview];
+}
--- a/cmake/feature-panels.cmake
+++ b/cmake/feature-panels.cmake
@@ -12,6 +12,10 @@ target_compile_definitions(browser-panels INTERFACE BROWSER_AVAILABLE)
 target_sources(obs-browser PRIVATE panel/browser-panel-client.hpp panel/browser-panel-internal.hpp
                                    panel/browser-panel.cpp panel/browser-panel-client.cpp)
 
+if(OS_MACOS)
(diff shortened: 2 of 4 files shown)


panel/browser-panel-macos.m
```diff
@@ -0,0 +1,15 @@
+#import <AppKit/AppKit.h>
+
+void RemoveNSViewFromSuperview(void *view)
```
PatTheMav, line 3: Usually it should be possible to cast the result of `GetWindowHandle()` directly to `NSView *`. And ObjC classes (even empty ones) can always be used for message passing, as even `nil` (but not `NULL`) accepts those.

So just checking for a null pointer should be sufficient.

gxalpha: > Usually it should be possible to cast the result of GetWindowHandle() directly to NSView *

It should be, but not in C++, right? Or is there a way to (opaquely) use Objective-C classes in (normal) C++?
My idea here was to keep Objective-C out of the C++ file, but also to avoid Objective-C++ in `browser-panel-macos.m(m)` (just because I'm personally not a big fan of ObjC++).

> ObjC classes (even empty ones) can always be used for message passing, as even nil (but not NULL) accepts those.
So just checking for a null pointer should be sufficient.

Interesting, that doesn't work for me.
For the original bug, the pointer isn't a nullptr, but its memory is zeroed. Sending the `removeFromSuperview` message on it crashes reliably, e.g. when closing the Twitch login panel. This is where the (existing) `(*((bool *)view))` comes from.
The `(!view)` is my guesswork for the new crash that I cannot reproduce in the first place.

PR #423, opened 2023-08-23
Add function to resize the browser source from javascript

exeldro:
### Description
Add function to resize the browser source from javascript

### Motivation and Context
Allows browser sources to be sized to exactly fit its contents.
This can be very useful for my Markdown Source plugin.

### How Has This Been Tested?
On windows 11 64 bit using Markdown Source with the following javascript:
```
	let size = document.body.getBoundingClientRect();
	window.obsstudio.setBrowserSize(Math.ceil(size.width), Math.ceil(size.height));
```

### Types of changes

### Checklist:


browser-client.cpp
```diff
@@ -198,6 +198,30 @@ bool BrowserClient::OnProcessMessageReceived(
...
+				height = (int)round(input_args->GetDouble(2));
+			if (width > 0 && height > 0) {
+				obs_data_t *s =
+					obs_source_get_settings(bs->source);
+				if (s &&
+				    (obs_data_get_int(s, "width") != width ||
+				     obs_data_get_int(s, "height") != height)) {
+					obs_data_set_int(s, "width", width);
+					obs_data_set_int(s, "height", height);
+					obs_source_update(bs->source, nullptr);
```
Lain-B, line 216: `bs->source` probably should be `s` just to be safe.

Lain-B (changes requested):
Would love to have more feedback from others, but otherwise I don't mind one way or another. Maybe @Warchamp7 has an opinion on this one?

But here's one super tiny nitpick.

browser-client.cpp
exeldro: I am not sure what you mean as `bs->source` is a `obs_source_t*` and `s` is a `obs_data_t*`.

gxalpha, line 216: I think this isn't really how `obs_source_update` is meant to be used.
Sure, you *can* do it like this, but really you're meant to pass a new `obs_data_t` object containing the new settings (in place of the `nullptr`), instead of modifying the object obtained by `obs_source_get_settings` directly.

And while you're at it, you should probably use the C++-RAII-types (`OBSData` / `OBSDataAutoRelease`) in C++-code.

Warchamp7:
What's the use case for adding a new obs-browser js binding over say using obs-websockets? Is it just the hurdle of connecting to obs-ws?

Ever since obs-ws was made a core plugin I'm generally against adding new obs-browser bindings to avoid maintaining essentially two browser source APIs.


At the very least, I think a better approach here would be a more abstract system that allows browser sources to dispatch() an event to OBS plugins. Then your plugin could handle the source resizing in response to that event.

exeldro:
The use case is my Markdown Source plugin. I want to me able to size the browser source to its contents provided by the markdown. I could hack a way by forcing obs-websocket enabled and getting the obs-websocket port and password from the settings and provide that to the the browser source, but I prefer a way that I don't have to force users to enable obs-websocket.

PR #429, opened 2023-11-17
Fix handling frontend JS events

### LINKED ISSUE #428: Some Events (addEventListener) are not functioning in OBS30+
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

30.0.0

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/OHn6774kHAQOU11w

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Events work

### Current Behavior

Most Events bound to addEventListener (on window) do not function/get called

The following still work/I've tested to still work:
- `obsSceneChanged`
- `obsSourceVisibleChanged`
- `obsSourceActiveChanged`

(Tested most of the other events and nothing else seems to work, most notabled stream/recording starting/started/stopping/stopped which are the key events I utilise)

### Steps to Reproduce

In a browser source use following example javascript code results in no console.log occuring indicating the event wasn't called.

```js

    window.addEventListener('obsStreamingStarting', function() {
        console.log('OBS: obsStreamingStarting');
    });
    window.addEventListener('obsStreamingStarted', function() {
        console.log('OBS: obsStreamingStarted');
    });
    window.addEventListener('obsStreamingStopping', function() {
        console.log('OBS: obsStreamingStopping');
    });
    window.addEventListener('obsStreamingStopped', function() {
        console.log('OBS: obsStreamingStopped');
    });
    window.addEventListener('obsRecordingStarting', function() { console.log('OBS: obsRecordingStarting'); });
    window.addEventListener('obsRecordingStarted', function() { console.log('OBS: obsRecordingStarted'); });
    window.addEventListener('obsRecordingStopping', function() { console.log('OBS: obsRecordingStopping'); });
    window.addEventListener('obsRecordingStopped', function() { console.log('OBS: obsRecordingStopped'); });
```

I tested this on

Control Level 2: Read Access to user information

and

Control Level 5: EVERYTHING (full access to OBS)

Under Control Level 5 the call to start/stop recording works via JavaScript so it seems that it is just the events side.

And/or using the UI buttons to start/stop recording/of streaming under control level 5 no console.log's occur/event emissions

I did not test _every_ function jsut the listed "broken" ones above and `obsSceneChanged`

RytoEX:
### Description

When handling frontend events, DispatchJSEvent expects JSON passed as a std::string. An unquoted string is not valid JSON. To get valid JSON, we can build a JSON object like in OBS_FRONTEND_EVENT_SCENE_CHANGED and pass the result of dump() to DispatchJSEvent.

When handling frontend events, DispatchJSEvent expects JSON passed as a std::string. An empty string will cause the event to never be dispatched. An empty nlohmann::json object converted to a std::string is represented by the string "null". Passing this exact string as the JSON string to DispatchJSEvent makes the events work. This is probably because an empty unquoted string by itself is not valid JSON. Two double-quotes with nothing inside them or the unquoted string "null" are considered valid JSON.

### Motivation and Context

Fixes #428 

### How Has This Been Tested?

Tested locally on Windows 11.

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-plugin.cpp
+++ b/obs-browser-plugin.cpp
@@ -549,55 +549,55 @@ static void handle_obs_frontend_event(enum obs_frontend_event event, void *)
 {
 	switch (event) {
 	case OBS_FRONTEND_EVENT_STREAMING_STARTING:
-		DispatchJSEvent("obsStreamingStarting", "");
+		DispatchJSEvent("obsStreamingStarting", "null");
 		break;
 	case OBS_FRONTEND_EVENT_STREAMING_STARTED:
-		DispatchJSEvent("obsStreamingStarted", "");
+		DispatchJSEvent("obsStreamingStarted", "null");
 		break;
 	case OBS_FRONTEND_EVENT_STREAMING_STOPPING:
-		DispatchJSEvent("obsStreamingStopping", "");
+		DispatchJSEvent("obsStreamingStopping", "null");
 		break;
 	case OBS_FRONTEND_EVENT_STREAMING_STOPPED:
-		DispatchJSEvent("obsStreamingStopped", "");
+		DispatchJSEvent("obsStreamingStopped", "null");
 		break;
 	case OBS_FRONTEND_EVENT_RECORDING_STARTING:
-		DispatchJSEvent("obsRecordingStarting", "");
+		DispatchJSEvent("obsRecordingStarting", "null");
 		break;
 	case OBS_FRONTEND_EVENT_RECORDING_STARTED:
-		DispatchJSEvent("obsRecordingStarted", "");
+		DispatchJSEvent("obsRecordingStarted", "null");
 		break;
 	case OBS_FRONTEND_EVENT_RECORDING_PAUSED:
-		DispatchJSEvent("obsRecordingPaused", "");
+		DispatchJSEvent("obsRecordingPaused", "null");
 		break;
 	case OBS_FRONTEND_EVENT_RECORDING_UNPAUSED:
-		DispatchJSEvent("obsRecordingUnpaused", "");
+		DispatchJSEvent("obsRecordingUnpaused", "null");
 		break;
 	case OBS_FRONTEND_EVENT_RECORDING_STOPPING:
-		DispatchJSEvent("obsRecordingStopping", "");
+		DispatchJSEvent("obsRecordingStopping", "null");
 		break;
 	case OBS_FRONTEND_EVENT_RECORDING_STOPPED:
-		DispatchJSEvent("obsRecordingStopped", "");
+		DispatchJSEvent("obsRecordingStopped", "null");
 		break;
 	case OBS_FRONTEND_EVENT_REPLAY_BUFFER_STARTING:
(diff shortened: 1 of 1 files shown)


WizardCM:
Thanks for digging into this. In my own testing I did track it down to the change away from json11, but couldn't dig down further.

I agree with the solution in https://github.com/obsproject/obs-browser/pull/429/commits/914368745e5ae70df699e5ba245d3b478eac9695 as it's the right change.

For the second commit, I'd like to propose an alternative:

```diff
diff --git a/browser-app.cpp b/browser-app.cpp
index 582b1c6e..71975ad0 100644
--- a/browser-app.cpp
+++ b/browser-app.cpp
@@ -328,8 +328,12 @@ bool BrowserApp::OnProcessMessageReceived(CefRefPtr<CefBrowser> browser,
                ExecuteJSFunction(browser, "onActiveChange", arguments);

        } else if (message->GetName() == "DispatchJSEvent") {
-               nlohmann::json payloadJson = nlohmann::json::parse(
-                       args->GetString(1).ToString(), nullptr, false);
+               CefString data = args->GetString(1);
+               if (data.empty())
+                       data = "null";
+
+               nlohmann::json payloadJson =
+                       nlohmann::json::parse(data.ToString(), nullptr, false);

                nlohmann::json wrapperJson;
                if (args->GetSize() > 1)
```

My method of tracking down exactly how best to solve this was by checking `payloadJson.is_discarded()` which returns `true` if the string passed to nlohmann is invalid JSON.

You are correct that we should be setting the JSON to `null`, as that is consistent with events pre-30 (verified in OBS 29.1). However, for overall readability, I'd like to stick with passing an empty string to `DispatchJSEvent`.

WizardCM (changes requested):
See my previous comment on current thoughts.

RytoEX:
> See my previous comment on current thoughts.

I'm a bit torn. On the one hand, your proposal catches all possible invocations of DispatchJSEvent. On the other hand, it does introduce some extra work on processing events (though it's probably negligible), and I personally prefer that we just pass valid JSON to DispatchJSEvent when calling it because I think it's more explicit, though I understand the appeal of catching unexpected cases like this. Ultimately, I don't have a strong argument for one proposal over the other.

BarryCarlyon:
I would not consider it a "breaking change" to change to a new behaviour in OBS 30. (passing an empty string)

Since the documentation for the events that (will now) have an empty string, people are not (or shouldn't) be expecting to have anything passed to them anyway on those events. And even if you are JSON parsing blinding you'd be wrapping in a try/catch in your browser source HTML.

So I'd throw a vote in for the new behaviour (empty string), then we can get this shipped in the OBS hotfix.

We just might need also consider a tweak to the documentation about nulls/empty strings _if_ you are trying to capture and parse a return value to the listener for an event where one isn't expected anyway. For "bad devs" (used loosley) a empty string is still as "falsey" as a null

RytoEX:
To prevent this from sitting longer, I'm going to merge this as-is, since this is the code I have already tested. If there are further changes we wish to make later, feel free to open a PR.

PR #431, opened 2024-01-10
Add procedure handler based API header

tytan652:
### Checklist:

- [ ] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [ ] All commit messages are properly formatted and commits squashed where appropriate.


PatTheMav:
I think the approach is correct and barring some concerns regarding the name (`PhQCefWidget` is a mouthful, `OBSBrowserWidget` would be more expressive, especially if this becomes a public API).

As for the API - how can we make this API available as part of `obs-frontend-api` without also exposing `obs-browser` publicly?

Because plugins need to be built out-of-tree, the browser API functions need to be available via the frontend API only, so it needs to be exposed there and needs to be compiled and linked without the browser plugin present.

tytan652:
> As for the API - how can we make this API available as part of `obs-frontend-api` without also exposing `obs-browser` publicly?

This design is meant to make it unrelated to `obs-frontend-api` (beside obs-browser internal using it).

> …the browser API functions need to be available via the frontend API only…

This API does not need/depend on the frontend API.

PatTheMav:
> > As for the API - how can we make this API available as part of `obs-frontend-api` without also exposing `obs-browser` publicly?
> 
> This design is meant to make it unrelated to `obs-frontend-api` (beside obs-browser internal using it).
> 
> > …the browser API functions need to be available via the frontend API only…
> 
> This API does not need/depend on the frontend API.

Got it. In that case this PR cannot supersede the PRs mentioned in the OP - the goal of at least one of those is to allow external plugins to create and manage their own browser docks.

As those plugins only link with `libobs` and `obs-frontend-api`, such functionality would need to be made available in the latter.

I still think it's worthwhile to have an API for internal use that does not require internal knowledge of the `obs-browser` plugin to interact with panels, but it's not solving the use case for external plugins then.

tytan652:
> Got it. In that case this PR cannot supersede the PRs mentioned in the OP - the goal of at least one of those is to allow external plugins to create and manage their own browser docks.

Beside the native flag issue.
This API can be used with the dock-related front-end API to make docks with a QCefWidget.

PatTheMav:
> > Got it. In that case this PR cannot supersede the PRs mentioned in the OP - the goal of at least one of those is to allow external plugins to create and manage their own browser docks.
> 
> Beside the native flag issue. This API can be used with the dock-related front-end API to make docks with a QCefWidget.

So with this merged, a leaner frontend API can be based on it you mean?

tytan652:
> So with this merged, a leaner frontend API can be based on it you mean?

1. Plugins creates everything to have a PhQCefWidget instance
2. Use `obs_frontend_add_dock_by_id("test", "Test", ph_qcef_widget.qwidget())`
3. A dock with browser widget is created

Like I said earlier, there is an because of the missing native flag but it's inside the Frontend API that the issue should be solved.

PatTheMav:
> > So with this merged, a leaner frontend API can be based on it you mean?
> 
> No need for new frontend API.
> 
> 1. Plugins creates everything to have a PhQCefWidget instance
> 2. Use `obs_frontend_add_dock_by_id("test", "Test", ph_qcef_widget.qwidget())`
> 3. A dock with browser widget is created
> 
> Like I said earlier, there is an because of the missing native flag but it's inside the Frontend API that the issue should be solved.

How can a plugin do that without access to the `PhQCefWidget` class? A plugin needs to be able to do that without access to `obs-browser`.

tytan652:
> How can a plugin do that without access to the `PhQCefWidget` class? A plugin needs to be able to do that without access to `obs-browser`. Or is that just a string identifier and the plugin needs to throw around a void pointer?

The header with PhCef* classes is public.

PatTheMav:
> > How can a plugin do that without access to the `PhQCefWidget` class? A plugin needs to be able to do that without access to `obs-browser`. Or is that just a string identifier and the plugin needs to throw around a void pointer?
> 
> The header with PhCef* classes is "public", no change in CMake was made to install it for now.

That's what I was hinting at - `obs-browser` is not allowed to install headers. Only `libobs` and `obs-frontend-api` are. If `PhCef*` needs to be made available, it needs to happen in `obs-frontend-api`. No other module is allowed to expose anything to plugins.

tytan652:
The header is not part of the obs-browser CMake module, for now it's an interface.

PatTheMav:
> The header is not part of the obs-browser CMake module, for now it's a standalone interface.

Ah my bad - that header is not part of this PR then but would be added in an associated obs-studio PR?

tytan652:
> Ah my bad - that header is not part of this PR then but would be added in an associated obs-studio PR?

`lib/obs-browser-api.hpp` is the header, it is in the PR and the API is made with procedure handler to avoid being dependent of the module. But obs-browser needs the header to have the API version macro.

tytan652:
The API header is not installed with other OBS Studio headers and I changed the *namespace* of the API classes (`Ph`-->`OBSBrowser`).

tytan652:
Update CMake part to match https://github.com/obsproject/obs-websocket/pull/1196 CMake as Pat approved it, like I said there https://github.com/obsproject/obs-websocket/pull/1196#discussion_r1468463474.

PR #434, opened 2024-02-04
Enable building with CEF 6261 & 6367 (Chromium 122 & 124)

WizardCM:
### Description

This allows compiling OBS Browser with Chromium 124-based CEF versions.

This also adds basic support for the new, official, shared textures API on both Windows and macOS.

Sources:
 - https://github.com/chromiumembedded/cef/blame/d3a483ef59f8f40c624967361e05edda413bbf5b/libcef_dll/ctocpp/browser_ctocpp.h#L57
 - https://bitbucket.org/chromiumembedded/cef/commits/f3b570c
 - https://bitbucket.org/chromiumembedded/cef/commits/260dd0ca2

Fun note; Remote debugging now requires `--remote-allow-origins=*` otherwise Remote Debugging immediately disconnects - even for localhost.

### Motivation and Context

At some point we'd like to upgrade to modern CEF. To do that, obs-browser needs to compile.

### How Has This Been Tested?

Launch OBS with a CEF 6261 or 6367 build. Have not tested downgrading CEF to our currently supported versions.

Only tested on Windows.

### Types of changes

### Checklist:

### DIFF
--- a/browser-client.cpp
+++ b/browser-client.cpp
@@ -373,7 +373,11 @@ void BrowserClient::UpdateExtraTexture()
 
 void BrowserClient::OnAcceleratedPaint(CefRefPtr<CefBrowser>,
 				       PaintElementType type, const RectList &,
+#if CHROME_VERSION_BUILD >= 6367
+				       const CefAcceleratedPaintInfo &info)
+#else
 				       void *shared_handle)
+#endif
 {
 	if (type != PET_VIEW) {
 		// TODO Overlay texture on top of bs->texture
@@ -384,7 +388,7 @@ void BrowserClient::OnAcceleratedPaint(CefRefPtr<CefBrowser>,
 		return;
 	}
 
-#ifndef _WIN32
+#if !defined(_WIN32) && CHROME_VERSION_BUILD < 6367
 	if (shared_handle == bs->last_handle)
 		return;
 #endif
@@ -399,12 +403,20 @@ void BrowserClient::OnAcceleratedPaint(CefRefPtr<CefBrowser>,
 		bs->texture = nullptr;
 	}
 
-#if defined(__APPLE__) && CHROME_VERSION_BUILD > 4183
+#if defined(__APPLE__) && CHROME_VERSION_BUILD > 6367
+	bs->texture = gs_texture_create_from_iosurface(
+		(IOSurfaceRef)(uintptr_t)info.shared_texture_io_surface);
+#elif defined(__APPLE__) && CHROME_VERSION_BUILD > 4183
 	bs->texture = gs_texture_create_from_iosurface(
 		(IOSurfaceRef)(uintptr_t)shared_handle);
 #elif defined(_WIN32) && CHROME_VERSION_BUILD > 4183
 	bs->texture =
+#if CHROME_VERSION_BUILD >= 6367
+		gs_texture_open_nt_shared(
+			(uint32_t)(uintptr_t)info.shared_texture_handle);
+#else
 		gs_texture_open_nt_shared((uint32_t)(uintptr_t)shared_handle);
+#endif
 	//if (bs->texture)
 	//	gs_texture_acquire_sync(bs->texture, 1, INFINITE);
 
@@ -415,7 +427,13 @@ void BrowserClient::OnAcceleratedPaint(CefRefPtr<CefBrowser>,
 	UpdateExtraTexture();
 	obs_leave_graphics();
 
+#if defined(__APPLE__) && CHROME_VERSION_BUILD >= 6367
+	bs->last_handle = info.shared_texture_io_surface;
+#elif CHROME_VERSION_BUILD >= 6367
+	bs->last_handle = info.shared_texture_handle;
+#else
 	bs->last_handle = shared_handle;
+#endif
 }
 
 #ifdef CEF_ON_ACCELERATED_PAINT2
--- a/obs-browser-page/obs-browser-page-main.cpp
+++ b/obs-browser-page/obs-browser-page-main.cpp
@@ -79,8 +79,10 @@ int CALLBACK WinMain(HINSTANCE, HINSTANCE, LPSTR, int)
 	std::thread shutdown_check;
 
 	CefMainArgs mainArgs(nullptr);
+#if CHROME_VERSION_BUILD < 5615
 	if (!SetHighDPIv2Scaling())
 		CefEnableHighDPISupport();
+#endif
 
 	CefRefPtr<CefCommandLine> command_line =
 		CefCommandLine::CreateCommandLine();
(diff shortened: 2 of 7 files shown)


jmickelonis:
Slightly off-topic, but you might be interested to know.
I built with 6261 on Ubuntu 23.10.  Browser sources and docks do show up, but calls to ExecuteJavascript fail silently without effect.  6167 works fine.

PatTheMav:
@WizardCM do you want to include the texture flip override for builds above 6261 in this PR as well (as discussed off-thread)?

As the texture shared with us is originally created for video encoding, it's pre-flipped already.

RytoEX:
> @WizardCM do you want to include the texture flip override for builds above 6261 in this PR as well (as discussed off-thread)?
> 
> As the texture shared with us is originally created for video encoding, it's pre-flipped already.

Gentle ping, @WizardCM ?

WizardCM:
@RytoEX @PatTheMav Per prior discussion, this PR now makes the necessary adjustments (including `flip`) to correctly run shared textures with CEF 6367. I have tested it on Windows 10 with the 6367 CEF build **only.** I've also kept the two commits separate for easier review, and they can realistically be combined for final merge.

browser-client.cpp
```diff
@@ -373,7 +373,11 @@ void BrowserClient::UpdateExtraTexture()
...
 void BrowserClient::OnAcceleratedPaint(CefRefPtr<CefBrowser>,
 				       PaintElementType type, const RectList &,
+#if CHROME_VERSION_BUILD >= 6367
+				       const CefAcceleratedPaintInfo &info)
+#else
 				       void *shared_handle)
+#endif
 {
 	if (type != PET_VIEW) {
 		// TODO Overlay texture on top of bs->texture
```
reitowo, line 383: Also left a word here. This should be removed >= 6367 as it has no guarantee on both macOS and Windows that same handle = same content. On Linux this handle is no more valid because of that complex native pixmap structure.

```
#ifndef _WIN32
	if (shared_handle == bs->last_handle)
		return;
#endif
```

browser-client.cpp
```diff
@@ -415,7 +427,11 @@ void BrowserClient::OnAcceleratedPaint(CefRefPtr<CefBrowser>,
 	UpdateExtraTexture();
 	obs_leave_graphics();
 
+#if CHROME_VERSION_BUILD >= 6367
```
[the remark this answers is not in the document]
RytoEX: Is this a theoretical or confirmed problem? If this is definitely creating an observable problem after building against the latest version of CEF, then we should address it now. If this is merely a theoretical problem that has not been confirmed, then I don't want to spend too much time dwelling on it. I'd rather make it possible to make builds against CEF 122/124+ and figure out what problems exist and fix them.

browser-client.cpp
RytoEX: As with the other one, is this a theoretical or confirmed problem?

browser-client.cpp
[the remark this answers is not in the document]
reitowo: Although IOSurface pointers could be valid for this as discussed with @PatTheMav , DXGI handles are not. I think we should just remove this line and remove `last_handle`. Re-importing every frame just cost 50~100us and I think we can omit it.

The problem is this is a real theoretical problem and we will not know if underlying chromium code changes behaviour. The suggested behaviour doesn't let us "cache".

Considering a frame skip could be rare and hard to notice, I think we should just disable such "optimization" and do the absolute right thing, prevent letting someone in the world wondering why frames are skipped.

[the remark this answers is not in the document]
RytoEX: A problem can't be both "real" and "theoretical". I'm asking if there's an observable failure/error when building with this code against CEF 124+. As it is right now, we cannot test obs-browser with CEF 124+ in any meaningful way without this PR, which is my concern at the moment. I'm not saying "don't ever fix this". I'm saying "does this need to be fixed before merging", because otherwise we can't meaningfully test CEF 124+ without specifically pulling this PR locally.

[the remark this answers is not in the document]
reitowo: Theoretically, yes. You can also anaylze the code and get the same conclusion. So I think we would better remove this >= 6367 to take no risk.

WizardCM:
Alright, commits have been updated. Tested on macOS, and `OnScheduleMessagePumpWork` changed somewhere between 5563 and 5938, so our code handles that case too. Tested on Linux, and no additional changes were needed on our side (obviously without shared textures).

This now builds and runs on Windows, macOS and Linux without issue, and the first 2 with shared textures. In my opinion this seems good to merge.

For now, I've left the code to set `bs->last_handle` but it's not read. I'll let others confirm what we want to do there.

reitowo:
Great work!

I forgot why, but is there a chance to import gbm in Linux? Seems OBS is the easiest user to make this work because it has a handy render pipeline.

browser-client.cpp
PatTheMav: In theory this should not be an issue on macOS as textures backed by IOSurfaces are "live" (as in: When whatever the IOSurface references is updated, the texture's visible contents are updated as well).

So if the handle hasn't changed, but the associated texture contents have, it would be reflected by OBS' own texture.

In my tests this was used by the old shared texture implementation heavily (we were able to jump out early because the IOSurface hadn't changed) but in my tests it never seems to happen in the new implementation (the IOSurface is always different, which suggests that every frame is represented by a new device texture?).

jpietek:
> Great work!
> 
> I forgot why, but is there a chance to import gbm in Linux? Seems OBS is the easiest user to make this work because it has a handy render pipeline.

I tried once (obs discord long discussion), but failed on Nvidia with some old for of yours. Would it makes sense to try again with 126 chromium?

RytoEX:
There are still be open feedback threads. They should be resolved before this can be merged.

browser-client.cpp
WizardCM: I have left this code path untouched and version gated to pre-patch, and verified it works on all platforms as-is.

obs-browser-page/obs-browser-page-main.cpp
```diff
@@ -79,8 +79,10 @@ int CALLBACK WinMain(HINSTANCE, HINSTANCE, LPSTR, int)
 	std::thread shutdown_check;
 
 	CefMainArgs mainArgs(nullptr);
+#if CHROME_VERSION_BUILD < 5615
 	if (!SetHighDPIv2Scaling())
```
PatTheMav, line 83: Is this handled differently on Windows with versions >= 5615, making this unnecesary?

WizardCM: Yes, the function is entirely removed and the behaviour is default instead.

browser-client.cpp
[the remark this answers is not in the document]
WizardCM: For now, I've left the code to set bs->last_handle but it's not read. I personally didn't see any strange behaviours either way, but would rather go the safer route for now.

reitowo (approved):

RytoEX (commented):
The CEF API changes look good to me. I haven't done a local test build of this yet.

RytoEX (approved):
I can confirm that I can build OBS Studio on Windows against both CEF 103 (5060 and 126 (6478) with these changes.

PR #436, opened 2024-03-21
cmake: Set minimum CEF version to 95

RytoEX:
### Description

We know that our minimum supported CEF version is 95, and we know that OBS (or obs-browser) will fail to build against anything older. Specify the minimum version required in CMake so this becomes a configure failure instead of a build failure.

### Motivation and Context

Fixes https://github.com/obsproject/obs-studio/issues/7963.

Actually fixing on the obs-studio side will require a submodule update.

### How Has This Been Tested?

I configured obs-studio on Windows 11 with CEF 103.

### Types of changes

### Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@ if(NOT ENABLE_BROWSER)
   return()
 endif()
 
-find_package(CEF REQUIRED)
+find_package(CEF REQUIRED 95)
 find_package(nlohmann_json REQUIRED)
 
 add_library(obs-browser MODULE)
--- a/cmake/legacy.cmake
+++ b/cmake/legacy.cmake
@@ -25,7 +25,7 @@ option(ENABLE_BROWSER_QT_LOOP "Enable running CEF on the main UI thread alongsid
 
 mark_as_advanced(ENABLE_BROWSER_LEGACY ENABLE_BROWSER_SHARED_TEXTURE ENABLE_BROWSER_PANELS ENABLE_BROWSER_QT_LOOP)
 
-find_package(CEF REQUIRED)
+find_package(CEF REQUIRED 95)
 
 if(NOT TARGET CEF::Wrapper)
   message(


PatTheMav:
Wouldn't it make more sense to specify an `EXACT` version match given that we only support a specific version of CEF at any time and probably don't want to officially support building with untested/unproven newer versions?

RytoEX:
> Wouldn't it make more sense to specify an `EXACT` version match given that we only support a specific version of CEF at any time and probably don't want to officially support building with untested/unproven newer versions?

Not to me, no. This is a narrowly scoped PR to simply address the lack of a _lower bound_ at configure time. CEF 95 is the oldest version our code will build against at this time.

Furthermore, I did not want to create a barrier to development for those implementing support for newer CEF versions by setting further version requirements. Right now, our code has ifdefs for multiple versions of CEF. Historically, we have found that not all CEF upgrades go perfectly, and we have needed to maintain compatibility with different versions on different platforms such that a single exact version requirement would not work.

PatTheMav:
> > Wouldn't it make more sense to specify an `EXACT` version match given that we only support a specific version of CEF at any time and probably don't want to officially support building with untested/unproven newer versions?
> 
> Not to me, no. This is a narrowly scoped PR to simply address the lack of a _lower bound_ at configure time. CEF 95 is the oldest version our code will build against at this time.
> 
> Furthermore, I did not want to create a barrier to development for those implementing support for newer CEF versions by setting further version requirements. Right now, our code has ifdefs for multiple versions of CEF. Historically, we have found that not all CEF upgrades go perfectly, and we have needed to maintain compatibility with different versions on different platforms such that a single exact version requirement would not work.

If we treat this as a minimum version, then our build system effectively states

> You can use any version more recent than and including v95

If this is not the case code-wise, then adding a single version guard is incomplete. _If_ we add a version guard, it needs to be correct. So we either state the exact version that we support or a range of versions which we do indeed support building with.

If different OS have different version requirements, they also need to be added explicitly (see also FFmpeg, for which we have different requirements on Linux).

I just don't see a good reason to add a version check, but then not make full use of it by having it specific enough to prohibit configuring an impossible build.

RytoEX:
> If we treat this as a minimum version, then our build system effectively states
> > You can use any version more recent than and including v95

I disagree. In my reading, it communicates:
> You cannot configure with CEF versions older than 95.

Anything else is an interpretation.

> If this is not the case code-wise, then adding a single version guard is incomplete. _If_ we add a version guard, it needs to be correct. So we either state the exact version that we support or a range of versions which we do indeed support building with.

Except this effectively _is_ the case, code-wise. The code is _known_ to fail below CEF 95. It is _unknown_ what happens above CEF 103, other than so far vague reports of "it builds okay" (for versions 104 and up till somewhat recently) or in _some_ cases "requires changes" (see #434). This PR is narrowly scoped to address the _known_ failure, not constrain against _unknowns_.

> If different OS have different version requirements, they also need to be added explicitly (see also FFmpeg, for which we have different requirements on Linux).

They currently do not. I am speaking historically as to why we cannot (or could not) universally all the time require _one_ specific version of CEF. We also do not require _exact_ versions or even _version ranges_ of FFmpeg right now. We have set minimums. This mirrors that. If we want to start providing version ranges or maximum versions via CMake, I consider that outside the scope of this PR. This is specifically addressing the lack of a minimum version requirement at configure time which we _know_ will fail at build time.

> I just don't see a good reason to add a version check, but then not make full use of it by having it specific enough to prohibit configuring an impossible build.

As mentioned, this is fixing a specific GitHub Issue that is currently open on obs-studio that is trivial to fix that prevents a _known bad configuration_ (anything under CEF 95). If we don't want this, then we might as well wash our hands of it and set no version requirement and simply say "you use our packaged version or use your own at your own risk".

In my opinion, this is better than what we have now, which is no version constraints at configure time at all. If we want to discuss the merits of version maximums or version ranges, I would prefer we do that after fixing a known reported bug that is trivial to solve.

PatTheMav:
It's not an interpretation, with the change you can configure a project with _any_ CEF version above CEF 95. CMake will not prohibit anyone from configuring the project with - say CEF 115 - and thus still allows one to create a "broken by default" project config _just like when providing CEF 94 without the change_.

_If_ we want to use CMake to enforce/check for incompatibility with a provided CEF version, then we should do it correctly and explicitly specify the upper and lower version ranges that our code supports,

RytoEX:
> with the change you can configure a project with _any_ CEF version above CEF 95

Without this change, you can configure the project with _any_ CEF version at all. That seems worse than this.

WizardCM (approved):
I think setting a minimum is plenty. I'm surprised it handles major version so cleanly (I haven't tested this myself). I am OK with this change personally - adding an upper boundary would not match any of our other dependencies.

RytoEX:
Discussed off-thread with @PatTheMav . This is fine as-is, but his points are well taken. If we want to set maximum versions for dependencies in CMake, and I am not opposed to doing so, we can approach that as a separate matter. If and when we do so, we'll simply have to be mindful that we need to update CMake when updating for new versions.

PR #439, opened 2024-04-11
Add function to signal the browser source from JavaScript

exeldro:
### Description
adds `void source_browser_signal(ptr source, string signal)` to the global signal handler
adds `void browser_signal(ptr source, string signal)` to the signal handler of the browser source
in javascript you can signal a signal handler like this: `window.obsstudio.signal("signal")`

### Motivation and Context
Allow JavaScript to send signals to plugins, without the need for an obs websocket connection
More generic solution replacing #423 
The use case is my Markdown Source plugin. I want to me able to size the browser source to its contents provided by the markdown. I could hack a way by forcing obs-websocket enabled and getting the obs-websocket port and password from the settings and provide that to the the browser source, but I prefer a way that I don't have to force users to enable obs-websocket.

### How Has This Been Tested?
On windows 11 with changes to the markdown source to allow it to resize it self

### Types of changes
- New feature (non-breaking change which adds functionality) 

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #441, opened 2024-04-22
Add function to call websocket request from JavaScript

exeldro:
### Description
Add function to call websocket request from JavaScript
This will require page permission full access

### Motivation and Context
This makes browser source pages not having to setup a websocket connection to call websocket requests.
Other alternative to #439 and #423

### How Has This Been Tested?
On windows 11

### Types of changes

### Checklist:

### DIFF
--- a/browser-client.cpp
+++ b/browser-client.cpp
@@ -19,6 +19,7 @@
 #include "browser-client.hpp"
 #include "obs-browser-source.hpp"
 #include "base64/base64.hpp"
+#include "..\obs-websocket\lib\obs-websocket-api.h"
 #include <nlohmann/json.hpp>
 #include <obs-frontend-api.h>
 #include <obs.hpp>
@@ -142,6 +143,28 @@ bool BrowserClient::OnProcessMessageReceived(
 			obs_frontend_start_virtualcam();
 		} else if (name == "stopVirtualcam") {
 			obs_frontend_stop_virtualcam();
+		} else if (name == "callWebsocketRequest") {
+			std::string request_type =
+				input_args->GetString(1).ToString();
+			std::string request_data_string =
+				input_args->GetString(2).ToString();
+			OBSDataAutoRelease request_data =
+				obs_data_create_from_json(
+					request_data_string.c_str());
+			struct obs_websocket_request_response *response =
+				obs_websocket_call_request(request_type.c_str(),
+							   request_data);
+			if (response) {
+				json = {{"code", response->status_code},
+					{"result",
+					 response->status_code == 100}};
+				if (response->response_data)
+					json["responseData"] =
+						response->response_data;
+				if (response->comment)
+					json["comment"] = response->comment;
+				obs_websocket_request_response_free(response);
+			}
 		}
 		[[fallthrough]];
 	case ControlLevel::Advanced:
--- a/README.md
+++ b/README.md
@@ -315,6 +315,31 @@ Permissions required: ALL
 window.obsstudio.stopVirtualcam()
 ```
 
+#### Call Websocket request
+Permissions required: ALL
+```js
+/**
+ * @typedef {Object} WebsocketResponse
+ * @property {number} code - RequestStatus code
+ * @property {bool} result - is true if the request resulted in Success. False if otherwise.
+ * @property {string} responseData - JSON string containing response data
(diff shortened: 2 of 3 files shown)


exeldro:
Rebased

browser-client.cpp
```diff
@@ -19,6 +19,7 @@
 #include "browser-client.hpp"
 #include "obs-browser-source.hpp"
 #include "base64/base64.hpp"
+#include "..\obs-websocket\lib\obs-websocket-api.h"
```
tytan652, line 22: This header must not be include this way.

exeldro: Added a copy of the header file now

browser-client.cpp
```diff
@@ -19,6 +19,7 @@
 #include "browser-client.hpp"
 #include "obs-browser-source.hpp"
 #include "base64/base64.hpp"
+#include "obs-websocket-api.h"
```
tytan652, line 22:
```suggestion
#include <obs-websocket-api.h>
```

obs-websocket-api.h
tytan652, line 1: To remove (https://github.com/obsproject/obs-browser/pull/454)

exeldro:
Used better way to include obs-websocket-api.h
Added an error result for when obs-websocket is not loaded

tytan652 (approved):
obs-websocket-api is now correctly included

RytoEX:
On this, I need some explanation or justification as to why this is needed. @Warchamp7 @WizardCM ?

steveseguin:
Just my take, RytoEx:

VDO.Ninja and SocialStream.Ninja make use of the browser source API extensively, but it's rather limited relative to the websocket API.

The browser has WebRTC data-channel support, which bypasses firewalls, so users can control their OBS remotely without proxy servers or opening up their firewall.

Rather than trying to enhance the browser source API, or having users try to enable/configure web-sockets, connection states, ports, and security, just having the websocket functionality be made available via an internal API makes life easier. 

The permissions are marked as ControlLevel::All:, so it's possible to offer end users who wish to enable it to have clear context of what permissions are being enabled.

I've not tested it myself yet, but having to avoid setting up websockets as an end user to obtain the same functionality via browser source integrations would increase adoption of this advanced functionality. 

> On this, I need some explanation or justification as to why this is needed. @Warchamp7 @WizardCM ?

Fenrirthviti:
I am still very wary of adding this, as it is currently my understanding this just provides a full websocket connection without any authorization/interaction from an end user, correct?

If not, what is the exact flow for this, as an end-user who is adding a browser source (or browser dock) that utilizes this? Screenshots or documentation on how this will function in practice will be helpful here.

steveseguin:
As [exeldro](https://github.com/exeldro) notes,

At the very least currently, a user would need to select FULL ACCESS TO OBS, as a Page Permission, when setting up their OBS Browser source.  This already gives a remote user control to start/stop/record and change scenes.

(I would not be opposed to additional confirmation windows, prompting a user to the dangers. )

image

In the case of VDO.Ninja, a user would further have to include &remote, as a URL parameter, however this is specific to VDO.Ninja.  Once activated, the remote user (an IRL streamer on mobile, for example), has remote control to change scenes of their OBS Studio.

For example:

image

Users on my end want much more control over their OBS remotely; layouts, transitions, better scene state access, control over sources.

Fenrirthviti:
Thanks.

I understand the use case, but I still feel like this isn't the right approach. Right now, even the "full control" option for browser sources is *very* limited in what it can read or access. Exposing websockets is providing, effectively, unfettered access to almost the entire OBS internal API. We have historically rejected other additions to the JS API for being too permissive/risky, and I feel like this is the most extreme case of that.

I don't know offhand what an alternate proposal here would be. I get that you want to make it easier for users to set up remote control of OBS from a source outside their network, but again, I feel like the barrier to getting that set up is very intentional already. There is no real strong communication to the end user that setting "full control" is that dangerous with this new addition. The stated examples of start/stop stream are fairly benign, not "read and exfiltrate every single thing in OBS" which this would allow.

This likely needs input and guidance from @Warchamp7 before it can proceed, but I'm just going to note I'm against merging this in the current state, and at the very least, this will need a lot more discussion, and should probably be an RFC first.

PR #444, opened 2024-06-25
Added An Example Template

DanielLester83:
### Description
Added An Example Template

### Motivation and Context
Education
Code sharing

### Types of changes
Added html code
Added readme file


Fenrirthviti:
Thank you for the submission, but I don't think the browser source needs an HTML tutorial shipped alongside it.

PR #447, opened 2024-08-01
Enable Qt message loop on Linux

tytan652:
### Description

Fixes https://github.com/obsproject/obs-studio/issues/7146 for normally *any* CEF build.

This fixes issues cause by building CEF with use_gtk=true which is the default.

### Motivation and Context

We built CEF with `use_gtk=false` to avoid having collision between Qt and CEF using the Glib event loop.
Using the Qt loop avoids this collision to ever happen.

### How Has This Been Tested?

Type (keyboard) in a custom browser dock, a crash should occur with `use_gtk=true` with no Qt loop.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/os-linux.cmake
+++ b/cmake/os-linux.cmake
@@ -1,5 +1,7 @@
 find_package(X11 REQUIRED)
 
+target_compile_definitions(obs-browser PRIVATE ENABLE_BROWSER_QT_LOOP)
+
 target_link_libraries(obs-browser PRIVATE CEF::Wrapper CEF::Library X11::X11)
 set_target_properties(obs-browser PROPERTIES BUILD_RPATH "$ORIGIN/" INSTALL_RPATH "$ORIGIN/")
 


kkartaltepe:
thought we tried this and it didnt work to prevent gtk widgets from murdering cef.

tytan652:
`use_gtk=false` was mainly added to fix https://github.com/obsproject/obs-studio/issues/7146 caused by CEF(/GTK) messing with Glib default loop which is used by Qt.

If this does not work in the long term, a MainMessageLoop (like MainMessageLoopMultithreadedGtk in cefclient) would need to be implemented to be sure that CEF does not use the GLib default context directly.

RytoEX:
Is this change compatible with CEF 103/5060 as well as CEF 126+?

tytan652:
Tested with https://cef-builds.spotifycdn.com/cef_binary_103.0.12%2Bg8eb56c7%2Bchromium-103.0.5060.134_linux64_minimal.tar.bz2 (so `use_gtk=true`)

And the keyboard crash does not occur, and look like it works.

WizardCM (approved):

RytoEX (approved):
Tentatively seems fine to me.

PR #453, opened 2024-09-11
Add hardware accel support under Linux

tytan652:
### Description

Requires:
- https://github.com/obsproject/obs-studio/pull/11219

Required by:
- https://github.com/obsproject/obs-studio/pull/11262

Add hwaccel support on Linux with CEF 6533 and later

> [!CAUTION]
> ~~**NVIDIA might need to be blacklisted due to their inconsistency on supporting GBM and incomplete DRI3.**~~
> ~~TLDR: Only latest generations with latest driver has a likelihood to work~~
> **NVIDIA driver has been blacklisted**
>
> Unlike Intel and AMD (so mainly Mesa), NVIDIA themself created a feature disparity between generation even in the same series of driver.
>
> The blacklist is indeed quite wide but I don't have to all the card series to pinpoint what works and what does not. Including that when it "works" it needs to work under Wayland and X11 (yes both, not only one).

### Motivation and Context

Just wanted to try to make it somehow work for fun, ended up making it for real…

### How Has This Been Tested?

Tested under Wayland and X11.

### Types of changes

### Checklist:

### DIFF
--- a/browser-client.cpp
+++ b/browser-client.cpp
@@ -30,6 +30,10 @@
 #include <IOSurface/IOSurface.h>
 #endif
 
+#if !defined(_WIN32) && !defined(__APPLE__)
+#include "drm-format.hpp"
+#endif
+
 inline bool BrowserClient::valid() const
 {
 	return !!bs && !bs->destroying;
@@ -360,6 +364,31 @@ void BrowserClient::OnAcceleratedPaint(CefRefPtr<CefBrowser>, PaintElementType t
 		return;
 	}
 
+#if !defined(_WIN32) && !defined(__APPLE__)
+	if (info.plane_count == 0)
+		return;
+
+	struct obs_cef_video_format format = obs_cef_format_from_cef_type(info.format);
+
+	if (format.gs_format == GS_UNKNOWN)
+		return;
+
+	uint32_t *strides = (uint32_t *)alloca(info.plane_count * sizeof(uint32_t));
+	uint32_t *offsets = (uint32_t *)alloca(info.plane_count * sizeof(uint32_t));
+	uint64_t *modifiers = (uint64_t *)alloca(info.plane_count * sizeof(uint64_t));
+	int *fds = (int *)alloca(info.plane_count * sizeof(int));
+
+	for (size_t i = 0; i < kAcceleratedPaintMaxPlanes; i++) {
+		auto *plane = &info.planes[i];
+
+		strides[i] = plane->stride;
+		offsets[i] = plane->offset;
+		fds[i] = plane->fd;
+
+		modifiers[i] = info.modifier;
+	}
+#endif
+
 #if !defined(_WIN32) && CHROME_VERSION_BUILD < 6367
 	if (shared_handle == bs->last_handle)
 		return;
@@ -389,17 +418,21 @@ void BrowserClient::OnAcceleratedPaint(CefRefPtr<CefBrowser>, PaintElementType t
 	//if (bs->texture)
 	//	gs_texture_acquire_sync(bs->texture, 1, INFINITE);
 
-#else
+#elif defined(_WIN32)
 	bs->texture = gs_texture_open_shared((uint32_t)(uintptr_t)shared_handle);
+#else
+	bs->texture = gs_texture_create_from_dmabuf(bs->width, bs->height, format.drm_format, format.gs_format,
+						    info.plane_count, fds, strides, offsets,
+						    info.modifier != DRM_FORMAT_MOD_INVALID ? modifiers : NULL);
 #endif
 	UpdateExtraTexture();
 	obs_leave_graphics();
 
 #if defined(__APPLE__) && CHROME_VERSION_BUILD >= 6367
 	bs->last_handle = info.shared_texture_io_surface;
-#elif CHROME_VERSION_BUILD >= 6367
+#elif defined(_WIN32) && CHROME_VERSION_BUILD >= 6367
 	bs->last_handle = info.shared_texture_handle;
-#else
+#elif defined(__APPLE__) || defined(_WIN32)
 	bs->last_handle = shared_handle;
 #endif
 }
--- a/browser-client.cpp
+++ b/browser-client.cpp
@@ -31,6 +31,8 @@
 #endif
 
 #if !defined(_WIN32) && !defined(__APPLE__)
+#include <obs-nix-platform.h>
+
 #include "drm-format.hpp"
 #endif
 
@@ -369,6 +371,7 @@ void BrowserClient::OnAcceleratedPaint(CefRefPtr<CefBrowser>, PaintElementType t
 		return;
 
 	struct obs_cef_video_format format = obs_cef_format_from_cef_type(info.format);
+	uint64_t modifier = info.modifier;
 
 	if (format.gs_format == GS_UNKNOWN)
 		return;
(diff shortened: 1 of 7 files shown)


browser-client.cpp
```diff
@@ -388,6 +395,42 @@ void BrowserClient::OnAcceleratedPaint(CefRefPtr<CefBrowser>,
...
+		(uint32_t *)alloca(info.plane_count * sizeof(uint32_t));
+	uint64_t *modifiers =
+		(uint64_t *)alloca(info.plane_count * sizeof(uint64_t));
+	int *fds = (int *)alloca(info.plane_count * sizeof(int));
+
+	/* NOTE: This a workarround under X11 where the modifier is always
+	   invalid where it can mean "no modifier" in Chromium's code. */
+	if (obs_get_nix_platform() == OBS_NIX_PLATFORM_X11_EGL &&
+	    modifier == DRM_FORMAT_MOD_INVALID)
+		modifier = DRM_FORMAT_MOD_NONE;
```
kkartaltepe, line 421: Do not use the deprecated name MOD_NONE, use MOD_LINEAR.

kkartaltepe (approved):
Seems fine overall, i really wish we could pin down the modifier issue on x11 to something in cef's codebase... at least then we could see if it ever got fixed in the future.

kkartaltepe:
Doesnt compile with `ENABLE_BROWSER_SHARED_TEXTURE` set to 0 because this is used to avoid compiling the drm-formats stuff, but they are still referenced. Just remove the conditionals and always build these functions.

kkartaltepe:
Maybe i'm missing something on how this was tested, but current master sets the BrowserHwaccel option to false so nothing here takes effect until https://github.com/obsproject/obs-studio/blob/master/UI/obs-app.cpp#L1268-L1278 is flipped.

kkartaltepe:
https://gist.github.com/kkartaltepe/ebf66527a38f188c8b07eb43b2d37de6 I dont remember if this happened as well on my original test months ago. But on my steamdeck tier setup there is significant rendering performance loss so rendering is twice as slow for a scene with 10 browser sources.

tytan652:
> Doesnt compile with `ENABLE_BROWSER_SHARED_TEXTURE` set to 0 because this is used to avoid compiling the drm-formats stuff, but they are still referenced. Just remove the conditionals and always build these functions.

`#if ENABLE_BROWSER_SHARED_TEXTURE` should have been `#ifdef ENABLE_BROWSER_SHARED_TEXTURE`, I fixed it.
Our relies on whether it's define or not, not its value.

> I dont remember if this happened as well on my original test months ago. But on my steamdeck tier setup there is significant rendering performance loss so rendering is twice as slow for a scene with 10 browser sources.

This is why we test, I don't know I would be able to reproduce…

tytan652:
Changes:
- Avoid linking libdrm but expose it's include directory since we only need `DRM_FORMAT_`
- Blacklist NVIDIA driver since it does not work well on both Wayland and X11

fatalgoth:
Just streamed for six hours with these patches, minus the NVIDIA blacklisting and it solved some framedrops when certain browser overlays would draw to the screen for me. Instead of a blacklist, can it be a warning or something? It worked really well for me.

browser-client.cpp
```diff
@@ -381,17 +418,21 @@ void BrowserClient::OnAcceleratedPaint(CefRefPtr<CefBrowser>, PaintElementType t
...
 	UpdateExtraTexture();
 	obs_leave_graphics();
 
 #if defined(__APPLE__) && CHROME_VERSION_BUILD >= 6367
 	bs->last_handle = info.shared_texture_io_surface;
-#elif CHROME_VERSION_BUILD >= 6367
+#elif defined(_WIN32) && CHROME_VERSION_BUILD >= 6367
 	bs->last_handle = info.shared_texture_handle;
-#else
+#elif defined(__APPLE__) && defined(_WIN32)
```
RytoEX, line 435: How would both `__APPLE__` and `_WIN32` be defined? Was this meant to be inverted?
```cpp
#elif !defined(__APPLE__) && !defined(_WIN32)
```

browser-client.cpp
```diff
@@ -381,17 +418,21 @@ void BrowserClient::OnAcceleratedPaint(CefRefPtr<CefBrowser>, PaintElementType t
 	//if (bs->texture)
 	//	gs_texture_acquire_sync(bs->texture, 1, INFINITE);
 
-#else
+#elif defined(__WIN32)
```
RytoEX, line 421:
```suggestion
#elif defined(_WIN32)
```

RytoEX (changes requested):

obs-browser-plugin.cpp
```diff
@@ -692,6 +701,28 @@ static void check_hwaccel_support(void)
...
+	obs_leave_graphics();
+
+	if (strstr(glVersion, "NVIDIA") != NULL) {
+		hwaccel = false;
+		blog(LOG_INFO, "[obs-browser]: "
+			       "Blacklisted driver "
+			       "detected, "
+			       "disabling browser "
+			       "source hardware "
+			       "acceleration.");
```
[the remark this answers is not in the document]
tytan652: It was based on Windows `check_hwaccel_support`, but I did put the string as one piece in the new push.

browser-client.cpp
tytan652: Old shared_texture code path only meant for Windows and macOS, so `&&` should have been `||`. Fixed.

drm-format.cpp
```diff
@@ -0,0 +1,59 @@
...
+	},
+	{
+		CEF_COLOR_TYPE_BGRA_8888,
+		DRM_FORMAT_ARGB8888,
+		GS_BGRA_UNORM,
+		"BGRA_8888",
+	},
+};
+
+#define N_SUPPORTED_FORMATS (sizeof(supported_formats) / sizeof(supported_formats[0]))
```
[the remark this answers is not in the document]
tytan652: Converted to constexpr

RytoEX:
> https://gist.github.com/kkartaltepe/ebf66527a38f188c8b07eb43b2d37de6 I dont remember if this happened as well on my original test months ago. But on my steamdeck tier setup there is significant rendering performance loss so rendering is twice as slow for a scene with 10 browser sources.

@kkartaltepe If the original test was done with CEF 103 and the test in October was done with CEF 126/127, then could this be the issue with Qt message loop (https://github.com/obsproject/obs-studio/issues/11485 et al)? Or is it a separate issue?

Would still like to know if this is good enough to merge, or if there's something else we need to address first.

drm-format.hpp
```diff
@@ -0,0 +1,21 @@
...
+#ifdef ENABLE_BROWSER_SHARED_TEXTURE
+
+struct obs_cef_video_format {
+	cef_color_type_t cef_type;
+	uint32_t drm_format;
+	enum gs_color_format gs_format;
+	const char *pretty_name;
+};
+
+bool obs_cef_all_drm_formats_supported();
```
GeorgesStavracas, line 17: I don't know if this is the same in C++, but in C this would have been:

```suggestion
bool obs_cef_all_drm_formats_supported(void);
```

GeorgesStavracas (changes requested):

drm-format.hpp
tytan652: Not controversial so I made the change.

kkartaltepe (approved):

kkartaltepe:
> > https://gist.github.com/kkartaltepe/ebf66527a38f188c8b07eb43b2d37de6 I dont remember if this happened as well on my original test months ago. But on my steamdeck tier setup there is significant rendering performance loss so rendering is twice as slow for a scene with 10 browser sources.
> 
> @kkartaltepe If the original test was done with CEF 103 and the test in October was done with CEF 126/127, then could this be the issue with Qt message loop ([obsproject/obs-studio#11485](https://github.com/obsproject/obs-studio/issues/11485) et al)? Or is it a separate issue?
> 
> Would still like to know if this is good enough to merge, or if there's something else we need to address first.

No I think its unrelated to that. When i profiled it, graphics stuff was just really slow.

RytoEX (approved):

PR #458, opened 2024-10-12
cmake: Fix CEF minimum version

RytoEX:
### Description

Apparently, the find_package signature is incorrect. The [version] component must come before the [REQUIRED] component. I mistakenly placed them in the reverse order when originally adding the minimum version.

https://cmake.org/cmake/help/v3.28/command/find_package.html#basic-signature

### Motivation and Context

Noticed this while reviewing another PR. Want to fix my past mistakes.

### How Has This Been Tested?

Tested locally on Windows 11.
Before:
```
-- Found CEF: C:/dev/obsproject/obs-studio/.deps/cef_binary_6533_windows_x64/Release/libcef.dll (found version "127.3.4")
```

After:
```
-- Found CEF: C:/dev/obsproject/obs-studio/.deps/cef_binary_6533_windows_x64/Release/libcef.dll (found suitable version "127.3.4", minimum required is "95")
```

### Types of changes

### Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ if(NOT ENABLE_BROWSER)
   return()
 endif()
 
-find_package(CEF REQUIRED 95)
+find_package(CEF 95 REQUIRED)
 find_package(nlohmann_json 3.11 REQUIRED)
 
 add_library(obs-browser MODULE)


WizardCM (approved):

PatTheMav (commented):
Does the module even compile with a Chromium 95 build?

Because this instruction says it does, but if that wasn't the case then it's not just the order of arguments that is wrong.

RytoEX:
I have no reason to believe it wouldn't build, since we still have the ifdefs in place. I haven't tried yet though.

PatTheMav:
> I have no reason to believe it wouldn't build, since we still have the ifdefs in place. I haven't tried yet though.

I wouldn't have expected it to either, but if we use a version guard, it should also use the correct version and not allow a solution/project be generated when a severely outdated CEF version is provided.

RytoEX:
> > I have no reason to believe it wouldn't build, since we still have the ifdefs in place. I haven't tried yet though.
> 
> I wouldn't have expected it to either, but if we use a version guard, it should also use the correct version and not allow a solution/project be generated when a severely outdated CEF version is provided.

I used a double negative. To clarify, I would expect it to build, because we still have ifdefs in place down to that version.

WizardCM:
Yes, currently OBS is supposed to still build with Chromium 95 if the user chooses to provide the necessary CEF build.

If we successfully release with Chromium 127, the intent is to drop Chromium 95 and below in favour of 103 & 127.

I have verified current master does build with 95.

PR #450, opened 2024-08-18
Wait on shutdown for docks to close on Linux ('simplest' solution)

WizardCM:
### Description

Fixes a 100% reproducible crash on shutdown when running Chromium 127.

This is likely the same crash users have been running into with current CEF with a lower hit rate.

This is a temporary/simplified solution to allow a quick CEF upgrade.

Ideally we should track open docks & count `OnBeforeClose` calls before fully closing, per CEF documentation.

> The [CefLifeSpanHandler::OnBeforeClose()](https://cef-builds.spotifycdn.com/docs/127.0/classCefLifeSpanHandler.html#af79cf13d98aacb59618dbbe273cce0fa) method will be called after [DoClose()](https://cef-builds.spotifycdn.com/docs/127.0/classCefLifeSpanHandler.html#a4867b3d97d879d9996dec2cf1daa483d) (if [DoClose()](https://cef-builds.spotifycdn.com/docs/127.0/classCefLifeSpanHandler.html#a4867b3d97d879d9996dec2cf1daa483d) is called) and immediately before the browser object is destroyed. **The application should only exit after [OnBeforeClose()](https://cef-builds.spotifycdn.com/docs/127.0/classCefLifeSpanHandler.html#af79cf13d98aacb59618dbbe273cce0fa) has been called for all existing browsers.**

The same fix does not seem to be needed for browser sources, likely because they're closed earlier in the shutdown flow.

### Motivation and Context

Crashes are bad, including on shutdown.

### How Has This Been Tested?

Ubuntu 24.04.

* 1 dock (Google):   waits 30ms
* 1 dock (YouTube): waits 100ms
* 2 docks (Google + YouTube)):   waits 150ms

Without this fix, using a DCHECK enabled CEF build to read the error message:

```
[0818/184348.915859:FATAL:alloy_browser_main.cc(417)] Check failed: global_request_context_->HasOneRef(). 
```

With additional logging in OBS, the culprit was identified as Qt not waiting long enough for browser docks to close.
Good browser dock close (such as closing a single dock with its X):
```
warning: QCefWidgetInternal::~closeBrowser ---- starting
warning: QCefWidgetInternal::~closeBrowser ---- destroyed
warning: QCefWidgetInternal::~closeBrowser ---- ended
warning: ~QCefBrowserClient() ---- destroyed
```
Bad browser dock close (closing the app while browser docks are loaded):
```
warning: QCefWidgetInternal::~closeBrowser ---- starting
[...]

PR #460, opened 2024-10-18
Wait on shutdown for docks to close on Linux (Event-based solution)

WizardCM:
### Description

Solution 1 was a no-go with the Qt event loop, which we'd prefer to use.

This is solution two, which listens for an event from CEF when the shutdown is occurring. At most, it's configured to wait 1 second per dock (this is usually more than enough time, but we could bump it to 2 or 3 if needed).

This also moves widget destruction to *after* Qt destruction, otherwise CEF can't notify Qt that it's ready to close.

Also Closes #459

### Motivation and Context

The original solution in #450 works fine when not using the Qt event loop, however any method of waiting with the Qt event loop causes all CEF events to go on hold too, resulting in a hang or crash.

### How Has This Been Tested?

On Linux, with two browser docks: YouTube and Google. The former is considered a "heavier" dock and takes longer to close.

Also successfully loaded the Twitch OAuth window.

Tested with our current Beta 1-shipping CEF build of 127.

With this PR, there is no crash or hang on shutdown.

Note: Have not tested "What's New" myself.

### Types of changes

### Checklist:

### DIFF
--- a/panel/browser-panel-client.cpp
+++ b/panel/browser-panel-client.cpp
@@ -213,6 +213,13 @@ bool QCefBrowserClient::OnBeforePopup(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame>
 	return true;
 }
 
+void QCefBrowserClient::OnBeforeClose(CefRefPtr<CefBrowser>)
+{
+	if (widget) {
+		emit widget->CloseSafely();
+	}
+}
+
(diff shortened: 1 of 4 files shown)


tytan652:
Tested on my machine, all the freezes issues are fixed.

RytoEX (approved):
Seems fine on Windows.

PR #463, opened 2024-11-02
Expand panel shutdown event listener to all platforms

WizardCM:
### Description

This expands the previously introduced shutdown loop check for browser docks to Windows and macOS.

As a reminder, this solves a potential crash on shutdown if the user has many docks or heavy docks.

As this solution has a timeout (max 1 second per dock), I am much more comfortable expanding its usage to all platforms compared to the original solution.

### Motivation and Context

I have a feeling that users on other platforms have been running into this issue, but because we don't provide CEF symbols (and because it occurs on shutdown) reports have been sporadic and/or lacking the necessary detail to track it down.

### How Has This Been Tested?

Close browser docks manually and on shutdown, experience no crash.

### Types of changes

### Checklist:

### DIFF
--- a/panel/browser-panel.cpp
+++ b/panel/browser-panel.cpp
@@ -166,7 +166,7 @@ void QCefWidgetInternal::closeBrowser()
 
 			cefBrowser->GetHost()->CloseBrowser(true);
 
-#if !defined(_WIN32) && !defined(__APPLE__) && CHROME_VERSION_BUILD >= 6533
+#if CHROME_VERSION_BUILD >= 6533
 			QEventLoop loop;
 
 			connect(this, &QCefWidgetInternal::readyToClose, &loop, &QEventLoop::quit);


RytoEX:
If possible, I would like to see some users on Windows and macOS with heavy browser panel usage test this.

RytoEX:
Since this is expanding https://github.com/obsproject/obs-browser/pull/460, and that was mostly dealing with Qt event loop (Qt message loop?) usage, which has since been reverted, should we instead be gating this on `ENABLE_BROWSER_QT_LOOP`, which is presently only enabled on macOS? Or is the intent/belief that this shutdown method is safer with or without the Qt event loop?

WizardCM:
On the CEF side, you're always supposed to wait until all browser windows are closed before tearing it down. We call `CefShutdown` (and close OBS) regardless of whether we're using the Qt event loop in obs-browser, so this solution is required on all platforms in both use cases, as it just makes Qt wait until CEF is (more) ready to be safely shut down before doing so.

PR #472, opened 2025-01-31
Revert "Enable Qt message loop on Linux"

tytan652:
### Description

Closes https://github.com/obsproject/obs-studio/issues/11485

This reverts commit 98d94a432264c1754e4dea1501a5ce36b04b1718.

### Motivation and Context

Performance issues were met by multiple users.

So back to `use_gtk=false` builds.

### How Has This Been Tested?

Used `use_gtk=false` for testing (even tested with #453) everything seem to work properly.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/os-linux.cmake
+++ b/cmake/os-linux.cmake
@@ -1,7 +1,5 @@
 find_package(X11 REQUIRED)
 
-target_compile_definitions(obs-browser PRIVATE ENABLE_BROWSER_QT_LOOP)
-
 target_link_libraries(obs-browser PRIVATE CEF::Wrapper CEF::Library X11::X11)
 set_target_properties(obs-browser PROPERTIES BUILD_RPATH "$ORIGIN/" INSTALL_RPATH "$ORIGIN/")
 


RytoEX:
cc @kkartaltepe @GeorgesStavracas

WizardCM (approved):

kkartaltepe (approved):

kkartaltepe:
Everything looks good with a new build of cef and this applied.

RytoEX (approved):
This seems fine.

Note: this _must_ be paired with new CEF builds that use `use_gtk=false`.

asahilina:
I'm curious, what's the issue when the CEF build has `use_gtk=true` with this revert/change?

tytan652:
Qt by default relies on Glib for its event loop, which completely goes wrong with the GTK inside CEF handle (e.g. https://github.com/obsproject/obs-studio/issues/7146).

We also normally should not need use_gtk or use_qt being enabled in our usecase, since both can lead to side effects.

tytan652:
I can give it to you, note that we enable H264 decoding in ours.

For the latest CEF 6533, I used  `GN_DEFINES="is_official_build=true use_sysroot=true use_partition_alloc_as_malloc=false symbol_level=1 is_cfi=false use_thin_lto=false proprietary_codecs=true ffmpeg_branding=Chrome chrome_pgo_phase=0 use_gtk=false"` and note that  we use the sysroot.

asahilina:
Thanks! Looks like the only relevant override there is `use_gtk=false`, everything else is already handled similarly at some other level in CEF/Fedora packaging.

PR #482, opened 2025-04-23
Add workaround for browser dock hanging

lexano-ivs:
### Description

Create workaround for https://bugreports.qt.io/browse/QTBUG-136165. 

### Motivation and Context
Upcoming OBS Studio 31.1 release upgraded from Qt 6.6.3 to Qt 6.8.3. This upgrade revealed an incompatibility with browser docks and Qt 6.8.3. Specifically, when multiple browser docks were open, restarting OBS would often result in all docks flashing very quickly and eventually OBS would hang. This issue was always triggered when Twitch docks were enabled, which requires the appropriate Twitch authorization tokens during the build. Thankfully the TEB beta program caught this issue before OBS 31.1-beta1 was built so we had time to investigate.

### How Has This Been Tested?
Local testing with and without the fix in place. The underlying issue is very reproducible and the workaround so far has been 100% reliable.

### Types of changes

### Checklist:

### DIFF
--- a/panel/browser-panel-client.cpp
+++ b/panel/browser-panel-client.cpp
@@ -226,6 +226,14 @@ void QCefBrowserClient::OnBeforeClose(CefRefPtr<CefBrowser>)
 
 bool QCefBrowserClient::OnSetFocus(CefRefPtr<CefBrowser>, CefFocusHandler::FocusSource source)
 {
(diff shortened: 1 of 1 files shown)


WizardCM (approved):

RytoEX (approved):
Unfortunate, but I'd rather not have OBS stuck in an infinite loop and freeze/hang/crash.

PR #483, opened 2025-04-30
Replace Glad usage with newer libobs graphics API

tytan652:
### Description

Replace Glad usage with newer libobs graphics API

https://github.com/obsproject/obs-studio/commit/abb8cdf0dad2267b59072b1b83da24e5fcd0b19f

### Motivation and Context

Cleaner to not rely on GL directly.

### How Has This Been Tested?

Yes with https://github.com/obsproject/obs-studio/pull/11262

### Types of changes

### Checklist:

### DIFF
--- a/cmake/os-linux.cmake
+++ b/cmake/os-linux.cmake
@@ -4,7 +4,7 @@ get_target_property(libdrm_include_directories Libdrm::Libdrm INTERFACE_INCLUDE_
 
 target_include_directories(obs-browser PRIVATE ${libdrm_include_directories})
 
-target_link_libraries(obs-browser PRIVATE CEF::Wrapper CEF::Library X11::X11 OBS::glad)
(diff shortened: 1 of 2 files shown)


RytoEX:
Do we think the new API will reliably return the driver version on the platforms we care about? See:
* https://github.com/obsproject/obs-studio/pull/12103

@kkartaltepe ?

tytan652:
> Do we think the new API will reliably return the driver version on the platforms we care about?

`gs_get_driver_version` is implement on Linux (the only concerned platform here) which is `glGetString(GL_VERSION)` on the OpenGL renderer.
It also removes the dependency on OpenGL if another renderer is made one day.
The mentioned PR adds checks that should be there, also this PR adds a NULL check.

kkartaltepe (approved):

kkartaltepe:
Looks good for linux, if this is supposed to work on windows im guessing it just wont.

RytoEX (approved):

PR #484, opened 2025-05-04
Prevent Windows 11 timer resolution throttling to fix audio distortion

### LINKED ISSUE #406: Audio Crackle when browser source is set to "Control audio via OBS" / New: Same for Media Sources with instant sounds
### Operating System Info

Windows 11

### Other OS

_No response_

### OBS Studio Version

27.2.4

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/yGAxWFqdh0G7zqMf

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

No audio crackle

### Current Behavior

Audio in browser source that have the option "Control audio via OBS" enabled have distorted/crackling sound when monitoring is enabled. Monitored sound will crackle but not the audio in the output.

### Steps to Reproduce

1. Add a browser source where you can play audio (SoundAlerts for example)
2. Enable "Control audio via OBS" and "Output and Monitoring"
3. Play a sound and hear it crackle (shorter ones seem to trigger it more)

I have this issue everytime, doesn't matter if only OBS is running or not. Maybe the new CEF version behaves differently on W11.

xtfo:
### Description

Fixes audio distortion when using the "Control audio via OBS" setting.

According to the SetProcessInformation API documentation:
"By default in Windows 11 if a window owning process becomes fully occluded, minimized, or otherwise non-visible to the end user, and non-audible, Windows may automatically ignore the timer resolution request and thus does not guarantee a higher resolution than the default system resolution."

Opt out of this behavior.

### Motivation and Context

Fixes #406

### How Has This Been Tested?

On Windows 11 and Windows 10.

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-page/obs-browser-page-main.cpp
+++ b/obs-browser-page/obs-browser-page-main.cpp
@@ -74,6 +74,13 @@ static void shutdown_check_thread(DWORD parent_pid, DWORD main_thread_id)
 
 int CALLBACK WinMain(HINSTANCE, HINSTANCE, LPSTR, int)
 {
+	PROCESS_POWER_THROTTLING_STATE PowerThrottling;
+	PowerThrottling.Version = PROCESS_POWER_THROTTLING_CURRENT_VERSION;
+	PowerThrottling.ControlMask = PROCESS_POWER_THROTTLING_IGNORE_TIMER_RESOLUTION;
+	PowerThrottling.StateMask = 0;
+
+	SetProcessInformation(GetCurrentProcess(), ProcessPowerThrottling, &PowerThrottling, sizeof(PowerThrottling));
+
 	std::thread shutdown_check;
 
 	CefMainArgs mainArgs(nullptr);


prgmitchell:
Fixes for me, video from 31.0.3 where I minimize a few seconds after starting the recording:

https://github.com/user-attachments/assets/ae361fb0-4b0e-495c-a259-07f65bf9803b

Video w/ this change, same scenario minimizing after a few seconds:

https://github.com/user-attachments/assets/1eeda0d6-45bf-4b74-8af8-79dc7e9ef5f6

WizardCM (approved):
Confirmed this fixes the issue on Windows 11, and the code matches the example code in Microsoft's documentation.

RytoEX (approved):
Looks fine. There are also `PROCESS_POWER_THROTTLING_EXECUTION_SPEED` and `ProcessMemoryPriority`, but I'm not aware of specific issues currently caused by not specifying those behaviors.

PR #489, opened 2025-06-20
Add check for Metal in render function

PatTheMav:
### Description
Adds an explicit check for the OpenGL renderer in the current render function to set the special `rect` effect only for that renderer.

### Motivation and Context
Metal does not require the "rect" effect and can use the default effect just like D3D11.

This change does not negatively affect OBS Studio running with OpenGL as its renderer, but unlocks use of browser sources for the Metal renderer.

### How Has This Been Tested?
Tested on macOS 15.

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -633,7 +633,14 @@ void BrowserSource::Render()
 
 	if (texture) {
 #ifdef __APPLE__
-		gs_effect_t *effect = obs_get_base_effect((hwaccel) ? OBS_EFFECT_DEFAULT_RECT : OBS_EFFECT_DEFAULT);
+		int type = gs_get_device_type();
+		gs_effect_t *effect;
+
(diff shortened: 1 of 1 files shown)


obs-browser-source.cpp
```diff
@@ -633,7 +633,14 @@ void BrowserSource::Render()
 
 	if (texture) {
 #ifdef __APPLE__
-		gs_effect_t *effect = obs_get_base_effect((hwaccel) ? OBS_EFFECT_DEFAULT_RECT : OBS_EFFECT_DEFAULT);
+		int type = gs_get_device_type();
```
RytoEX, line 636: It might be more efficient to store the device type on initialization than calling this here in `Render`, but perhaps that can be changed later if there's any measurable impact.

RytoEX (approved):
Still approved, just commenting on the placement/timing of `gs_get_device_type()`.

PR #493, opened 2025-07-26
Fix stack smash crashes on Linux

asahilina:
### Description

See: https://github.com/chromiumembedded/cef/issues/3912
Fedora bug: https://bugzilla.redhat.com/show_bug.cgi?id=2379591

### How Has This Been Tested?

Local test build

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-page/obs-browser-page-main.cpp
+++ b/obs-browser-page/obs-browser-page-main.cpp
@@ -102,6 +102,9 @@ int CALLBACK WinMain(HINSTANCE, HINSTANCE, LPSTR, int)
 	}
 
 #else
+#if defined(NO_STACK_PROTECTOR)
+NO_STACK_PROTECTOR
+#endif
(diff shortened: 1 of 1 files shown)


kkartaltepe (approved):

tytan652:
It's not exclusive to 31.1 (earlier version have the issue) and will require to backport changes to CEF 127 to have `NO_STACK_PROTECTOR` defined.
So in my opinion it can wait until a future major/minor release.

asahilina:
The main issue is noise for distros that have coredumps/reporting (like Fedora). Other than that it seems to be harmless since the crashes only happen on process exit.

For the fedora RPMs I will push this patch in so it's fixed there, since otherwise users will get abrt popups and file bugs with us (I already got one). I can't say about the effect for other distros or through the official Flatpak. If this has been going on for a while and nobody has complained about it yet, maybe it's fine to wait?

Issue #73, opened 2017-04-15
Missing JS Function: getVisibilityState


SReject:
Would be nice if there was a function similar to `getCurrentScene` that would return the visibility state.

```js
// register the callback
window.obsstudio.getVisibilityState(function callback(state) {
    // Called after the BrowerSource's visibility state is retrieved
});
```

zavitax:
What's wrong with `document.hidden` and the rest of the Document Visibility API?
https://www.html5rocks.com/en/tutorials/pagevisibility/intro/

WizardCM:
Just tested this myself - document.hidden definitely seems to do the job.

PR #495, opened 2025-08-16
Add get/set source visibility functions

fauxnik:
### Description

This exposes a javascript function for determining the visiblity of arbitrary OBS sources and another for modifying said visibility. These functions are available to browser sources with the requisite permissions: ReadUser or higher for `getSourceVisible`, Advanced or higher for `setSourceVisible`.

### Motivation and Context

Excerpt from https://github.com/obsproject/obs-browser/issues/73#issuecomment-3190645799:

> I've found myself surprised to find that there's no `setVisibilityState` or `setSourceVisibility`. I was hoping to hide the browser source when an html video element finishes playing so that the source's hide transition can be triggered, and the automation software I use can later show the source again with a new URL.
> 
> I could see someone arguing that `getSourceVisibility` would make sense too, as the visibility of another source/non-browser source can't be determined [using the browser's Document Visibility API].

### How Has This Been Tested?

I'm still working on figuring out how to build _obs-browser_, so the change hasn't been tested yet. I'd appreciate any tips you have regarding this as it appears the _obs-studio_ build process has changed since the last time the _obs-browser_ build documentation was updated.

### Types of changes

- Documentation (a change to documentation pages)

### Checklist:

- [ ] The code has been tested. ← working on this, see above
- [x] I have included updates to all appropriate documentation (that I'm aware of).

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,24 @@ window.obsstudio.getCurrentTransition(function (transition) {
 })
 ```
 
+#### Get source visibility
+Permissions required: READ_USER
+```js
+/**
+ * @callback VisibilityCallback
+ * @param {boolean} visibility - True -> visible, False -> hidden
+ */
+
+/**
+ * @param {string} scene - Name of the scene
+ * @param {string} source - Name of the source
+ * @param {VisibilityCallback} cb - The callback that receives the current visibility.
+ */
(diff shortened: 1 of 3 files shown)


Fenrirthviti:
We have previously rejected additions to the JS API as it's not really something we feel is sustainable long-term for the project.

Is there a reason that you cannot use the websockets API for this?

PR #500, opened 2025-08-23
add getSceneItemList, getSceneItemTransform & getBrowserInputSettings…

deemonrider:
This is a follow up to: https://github.com/obsproject/obs-browser/pull/499 as the previous approach failed when duplicating a browser source reference.
This time I tried to stay with the implementation more closely with how the WebSocket handles it and giving developers the same flexibility.

### Description
This new feature adds:
- getSceneItemList
image

- getSceneItemTransform
image

- getBrowserInputSettings
image

I decided to bundle these three functions into a single PR since they are closely related:
- First, you can call getSceneItemList to retrieve all items in a scene.
- Next, use getBrowserInputSettings to identify which of those items are relevant.
- Finally, you can fetch the transformation details for each item individually using its unique ID.

### How Has This Been Tested?
Environment: Windows 11 x64, OBS Studio 31.1.2 (build 104)

- I've used the buildin obs debugging tool to access the console and compare the output to the settings I've changed of the browser source and then called in the console:
- I've tested the permission behavior handling if the user removes all permissions of the browser source -> returns null like the other functions.
- I've duplicated a browser source so I have two times the object in my sources to ensure this is working now aswell.

### Types of changes
New feature (non-breaking change which adds functionality)
Documentation (a change to documentation pages)

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,99 @@ window.obsstudio.getScenes(function (scenes) {
 })
 ```
 
+#### Get scene item list
+Permissions required: READ_USER
+```js
+/**
+ * @typedef {Object} SceneItem
+ * @property {number} id - unique scene item ID
+ * @property {string} name - name of the source
+ * @property {boolean} visible - whether the item is visible
+ * @property {boolean} locked - whether the item is locked
+ * @property {string} type - the source type
+ */
+
+/**
+ * @callback SceneItemListCallback
+ * @param {SceneItem[]} sceneItems
+ */
+
+/**
+ * @param {SceneItemListCallback} cb - The callback that receives the list of scene items in the current scene.
+ */
+window.obsstudio.getSceneItemList(function(sceneItems) {
+    console.log(sceneItems)
+})
+```
+
+#### Get scene item transform
+Permissions required: READ_USER
+```js
+/**
+ * @typedef {Object} Transform
+ * @property {Object} position - position of the item
+ * @property {number} position.x - x coordinate
+ * @property {number} position.y - y coordinate
+ * @property {number} rotation - rotation in degrees
+ * @property {Object} scale - scale of the item
+ * @property {number} scale.x - x scale factor
+ * @property {number} scale.y - y scale factor
+ * @property {number} alignment - alignment flags
+ * @property {number} boundsType - bounds type
+ * @property {number} boundsAlignment - bounds alignment
+ * @property {Object} bounds - bounds of the item
+ * @property {number} bounds.x - bounds width
+ * @property {number} bounds.y - bounds height
+ * @property {Object} crop - crop settings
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -166,6 +166,99 @@ window.obsstudio.getScenes(function (scenes) {
 })
 ```
 
+#### Get scene item list
+Permissions required: READ_USER
+```js
+/**
+ * @typedef {Object} SceneItem
```
MattIPv4: Would love to see transform (or at least position) information included for these, if that would be possible?

jonaswagern:
I would love to have this also. Bump

paul-lrr:
I see that this change allows a browser source to get a list of **all** the items in a scene, but how would it get its **own** name and/or sceneItemId? From what I understand, it needs that Id in order to get its own info from `getSceneItemTransform` and `getBrowserInputSettings`

deemonrider:
> I see that this change allows a browser source to get a list of **all** the items in a scene, but how would it get its **own** name and/or sceneItemId? From what I understand, it needs that Id in order to get its own info from `getSceneItemTransform` and `getBrowserInputSettings`


Call `getSceneItemList()` to get all items in the scene.

Iterate over them and call `getBrowserInputSettings(sceneItemId)` on each — the returned settings include the url property, which you can compare against your own window.location.href to identify yourself.

(There can be a scenario where a user duplicates a browser source in the same scene where you would end up finding twice or more sources with the same href - but that could and should also be handled easily, if you detect your own href is twice in the scene list -> show error message to the user. If its onces in the lists - its yours. Usually you also don't want to have users add your browser sources more than once per scene - so you can even prevent them to do that with this update.)

paul-lrr:
> Call `getSceneItemList()` to get all items in the scene.
> 
> Iterate over them and call `getBrowserInputSettings(sceneItemId)` on each — the returned settings include the url property, which you can compare against your own window.location.href to identify yourself.
> 
> (There can be a scenario where a user duplicates a browser source in the same scene where you would end up finding twice or more sources with the same href - but that could and should also be handled easily, if you detect your own href is twice in the scene list -> show error message to the user. If its onces in the lists - its yours. Usually you also don't want to have users add your browser sources more than once per scene - so you can even prevent them to do that with this update.)

Ah, ok. so similar to how you would do it with obs-websocket. I was asking because a browser source having easy access to it's own name/source ID is something I've been hoping could be added to obs-browser for a long time. It would be very useful to have a source able to manipulate itself with the new functions in this PR or via obs-websocket without having to go through and check all the urls. 

FWIW, I do actually sometimes find it useful to have multiple browser sources in a scene that are separate sources but have the same url. For instance, having sources of the same page but with a different browser size to use in different circumstances or having sources of the same url using query strings that may currently be the same but have the option to be different.

PR #517, opened 2026-03-07
Update to C++20

hoshinolina:
### Description

Current CEF requires compiling as C++20.

### Motivation and Context

Fails to build with CEF 145.

### How Has This Been Tested?

Tested as local build, testing in Fedora shortly.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/os-linux.cmake
+++ b/cmake/os-linux.cmake
@@ -19,6 +19,7 @@ target_sources(
 target_include_directories(browser-helper PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/deps"
                                                   "${CMAKE_CURRENT_SOURCE_DIR}/obs-browser-page")
 
+target_compile_features(browser-helper PRIVATE cxx_std_20)
 target_link_libraries(browser-helper PRIVATE CEF::Wrapper CEF::Library)
 
 target_sources(obs-browser PRIVATE deps/ip-string-posix.cpp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,7 +42,7 @@ target_sources(
 
 target_include_directories(obs-browser PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/deps")
 
-target_compile_features(obs-browser PRIVATE cxx_std_17)
+target_compile_features(obs-browser PRIVATE cxx_std_20)
 target_link_libraries(obs-browser PRIVATE OBS::libobs OBS::frontend-api OBS::websocket-api nlohmann_json::nlohmann_json)
 
 if(OS_WINDOWS)
--- a/cmake/os-macos.cmake
+++ b/cmake/os-macos.cmake
@@ -35,6 +35,8 @@ foreach(helper IN LISTS helper_suffixes)
 
   target_compile_definitions(${target_name} PRIVATE ENABLE_BROWSER_SHARED_TEXTURE)
 
+  target_compile_features(${target_name} PRIVATE cxx_std_20)
+
   if(CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 14.0.3)
     target_compile_options(${target_name} PRIVATE -Wno-error=unqualified-std-cast-call)
   endif()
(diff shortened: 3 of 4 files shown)


RytoEX (changes requested):
Fix the merge conflict caused by https://github.com/obsproject/obs-browser/pull/518.

cmake/os-linux.cmake
```diff
@@ -20,6 +20,7 @@ target_include_directories(browser-helper PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/d
                                                   "${CMAKE_CURRENT_SOURCE_DIR}/obs-browser-page")
 
 target_link_libraries(browser-helper PRIVATE CEF::Wrapper CEF::Library)
+target_compile_features(browser-helper PRIVATE cxx_std_20)
```
RytoEX: Super nit, but swap the order of these lines:

```suggestion
target_compile_features(browser-helper PRIVATE cxx_std_20)
target_link_libraries(browser-helper PRIVATE CEF::Wrapper CEF::Library)
```

RytoEX (changes requested):
Super nit, but looks fine otherwise.

PatTheMav (commented):
I'm on board with increasing the language level to C++20, however that change should affect all targets (and not just the Linux ones).

`obs-browser` is not a "standalone module" (I removed CMake code to that end IIRC) and thus should always inherit what the main project sets up as (via `CMAKE_CXX_STANDARD`).

And moving the project to C++20 is indeed in the works already, so the issue will be solved automatically through that.

hoshinolina:
FWIW this patch is needed to build OBS with current CEF today. I'm not a CMake nor C++ expert, but building just the browser bits as C++20 (which is what this PR does) seems to work fine (and this patch is what Fedora is shipping).

WizardCM (approved):
I can confirm updating obs-browser (or parent) is required for CEF 145+ compatibility, and that this PR on its own works correctly on Linux and Windows.

Approving with the understanding that this change could either be merged via this PR, or via a PR to the obs-studio repository itself (per Pat's comment)

out:
This patch is also needed on macOS, but it misses the macOS helper targets.

I hit this building #523 against CEF 147 on macOS arm64. Bumping `obs-browser` alone isn't enough there: `cmake/os-macos.cmake` creates four separate helper executables (`browser-helper`, `browser-helper_gpu`, `browser-helper_plugin`, `browser-helper_renderer`) in a loop, and they compile `browser-app.cpp` / `obs-browser-page-main.cpp` — so they pull in the CEF headers too. They set no standard of their own, so they inherit obs-studio's global `set(CMAKE_CXX_STANDARD 17)` and fail:

```
In file included from cef/include/cef_base.h:42:
In file included from cef/include/internal/cef_ptr.h:37:
In file included from cef/include/base/cef_ref_counted.h:54:
cef/include/base/cef_scoped_refptr.h:101:16: error: no member named 'same_as' in namespace 'std'
  101 |   return !std::same_as<std::decay_t<decltype(GetRefCountPreference<T>())>,
      |                ^~~~~~~
cef/include/base/cef_scoped_refptr.h:127:22: error: no member named 'derived_from' in namespace 'std'
cef/include/base/cef_scoped_refptr.h:296:19: error: no template named 'convertible_to' in namespace 'std'
```

The equivalent of the `os-linux.cmake` change, inside the `foreach(helper IN LISTS helper_suffixes)` loop:

```diff
--- a/cmake/os-macos.cmake
+++ b/cmake/os-macos.cmake
   target_compile_definitions(${target_name} PRIVATE ENABLE_BROWSER_SHARED_TEXTURE)
+
+  target_compile_features(${target_name} PRIVATE cxx_std_20)
```

With that plus this PR's existing changes, obs-browser builds and runs on macOS against CEF 147 — browser sources render and Chrome Runtime initialises correctly.

For context on why the helpers need explicit opt-in: obs-studio's C++20 groundwork has already landed (obsproject/obs-studio#12630 and obsproject/obs-studio#13320), but `cmake/common/compiler_common.cmake` still sets `CMAKE_CXX_STANDARD 17` globally — so per-target `target_compile_features` is currently the only thing raising the standard, and any target that doesn't ask for it silently stays on C++17.

One question I can't answer from here: `cmake/os-windows.cmake` defines `obs-browser-helper` as its own executable and likewise sets no standard, so it looks like it would need the same line. I have no Windows machine to verify that on, so I may be missing something that covers it there.

### Environment

| | |
|---|---|
| OS | macOS 26.5.2, arm64 |
| Toolchain | Xcode 26.6 (17F113), Apple clang 21.0.0, CMake 4.1.2 |
| CEF | `147.0.14+g76d2442+chromium-147.0.7727.138` (macosarm64) |
| obs-studio | `0218aa658` |

RytoEX (approved):
I've added additional changes to get this to build with CEF 148/7778 on macOS and Windows. We're going to go with this until we can fully update obs-studio to C++20.

PR #518, opened 2026-04-10
Replace implicit lamda captures for C++20

RytoEX:
### Description

Implicit capture of 'this' with a capture default of '=' is deprecated in C++20. While the standard recommends replacing '=' with '=, this', that syntax is not C++17 compatible, and we will likely need some transition period where obs-studio and its modules are still built using C++17.

### Motivation and Context

Want obs-studio and its submodules to build with C++20 some day.

### How Has This Been Tested?

Compiled and tested on Windows 11.

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -171,7 +171,7 @@ void BrowserSource::ExecuteOnBrowser(BrowserFunc func, bool async)
 #ifdef ENABLE_BROWSER_QT_LOOP
 			QueueBrowserTask(cefBrowser, func);
 #else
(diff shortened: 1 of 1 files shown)


PatTheMav (approved):
My initial reaction is that "explicit capture lists" are actually great (explicit is better than implicit), though too long lists might become harder to parse.

I think in that case the same applies as with function signatures: If you design it in such a way that it requires ~10 arguments, chances are that those arguments might be better placed in a structure and that structure being passed along as the argument (or context data).

PR #519, opened 2026-04-14
Do not try to create the browser before init

hoshinolina:
### Description

Should fix the random asserts on startup (not related to the recent consistent startup failures), I think.

### Motivation and Context
Fixes a race.

### How Has This Been Tested?
Quick smoke test.

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -43,6 +43,7 @@
 
 using namespace std;
 
+extern os_event_t *cef_started_event;
 extern bool QueueCEFTask(std::function<void()> task);
 
 static mutex browser_list_mutex;
@@ -549,6 +550,9 @@ void BrowserSource::Update(obs_data_t *settings)
 
 void BrowserSource::Tick()
 {
+	if (os_event_try(cef_started_event) != 0)
+		return;
+
 	if (create_browser && CreateBrowser())
 		create_browser = false;
 #if defined(ENABLE_BROWSER_SHARED_TEXTURE)


notr1ch (approved):

PatTheMav:
What's the actual root cause of the race? At first glance the "browser manager thread" seems like a culprit, because it decouples the actual initialisation of CEF from the main thread on Windows and Linux (but not on macOS), so CEF will only be initialised "eventually".

On macOS `BrowserInit` always finishes before `obs_browser_initialize` returns, so I guess we don't face the same issue there?

hoshinolina:
Yes, it's that. In particular, the graphics thread will call tick on the source once it's added to the scene, and if CEF init isn't complete yet by then bad things happen.

PatTheMav (commented):
Annoying that it's necessary, but seems like the most elegant fix for this race condition.

Please amend the commit description per our contribution guidelines to explain the "why" of the fix (explaining the race condition, it's root cause, and why this fixes it).

hoshinolina:
Done.

obs-browser-source.cpp
```diff
@@ -549,6 +550,9 @@ void BrowserSource::Update(obs_data_t *settings)
 
 void BrowserSource::Tick()
 {
+	if (os_event_try(cef_started_event) != 0)
+		return;
```
PatTheMav, line 554:
```suggestion
	if (os_event_try(cef_started_event) != 0) {
		return;
	}
```
Sorry, should've picked up on this in my earlier preview. Per our code style guidelines, except for C code, we prefer all conditional and loop blocks to be explicitly enclosed with curlies.

hoshinolina: I mean... sure but... nothing else in the file does this, including the rest of the function.

I think if this is desired, someone should add `InsertBraces: true` to .clang-format and fix the whole codebase. Enforcing things like this in review is essentially a waste of everyone's time when clang-format can do it for you.

RytoEX: Historically, we have avoided this because it would be disruptive to all open PRs. However, we have decided that there is no good time for this, so we will be enforcing braces via clang-format in the near future.

Warchamp7 (approved):

PatTheMav (approved):

RytoEX (approved):

PR #523, opened 2026-05-24
Enable support for CEF 6613+ (Chrome 128 to 150) and Chrome Runtime

WizardCM:
### Description

We made it! After a lot of work, burnout, and testing, support for CEF 6613 (Chrome 128) and the Chrome Runtime is ready to be reviewed!

This (finally) gives us the ability to safely migrate to the Chrome Runtime, the replacement to the Alloy Runtime that CEF has been using forever. CEF/Chrome versions from 6613/128 to 7871/150 (yes, current stable) are now fully supported and work out of the box.

> [!TIP]
>  * To use this with Chrome 145+ https://github.com/obsproject/obs-browser/pull/517 is required.

Test builds of CEF can be downloaded from https://cef-builds.spotifycdn.com/index.html (use Minimal, Current Stable)

---

Add a dummy Browser Client to introduce limitations into the Chrome UI windows that can be opened in uncontrolled edge cases. Chrome UI windows are unmanaged by default. By returning something other than null in `GetDefaultClient()` we can lock it down in various ways. https://magpcss.org/ceforum/viewtopic.php?f=6&t=19898

`OnBeforePopup`, `OnOpenURLFromTab` in the Dummy Client can probably be removed as they don't get called. There's no direct way to stop the Chrome UI window from being opened, so instead close it immediately. However, I've kept them just in case.

Disable Various Chrome Settings in custom docks & browser sources via `SetPreferences`. For more, check [`chrome/common/pref_names.h`](https://github.com/chromium/chromium/blob/main/chrome/common/pref_names.h). We should actively keep an eye on this in future when updating CEF.

Chrome's default error display is now used by CEF, so on top of the existing override in browser docks, a similar override has been added to browser sources which just redirects to about:blank.

* https://github.com/chromiumembedded/cef/issues/3852

Also blocks
 - Chrome Extensions, as they are largely untested and unpredictable
 - MediaRouter, which provides Chrome's Cast.. functionality
 - CalculateNativeWinOcclusion, which lowers FPS for hidden pages
 - LiveCaption, which provides automatic captions
 - DocumentPictureInPictureAPI, which provides PiP widgets (YouTube)

A data migration function is included. It moves some directories into a new 'Default' profile subdirectory, and performs the necessary renaming of certain files to ensure the new Profile loads. This is necessary for user's data to be maintained, as the Alloy runtime used a bit of a custom structure which has been discarded in favour of Chrome Runtime's. CEF does not provide a migration path.

* https://github.com/chromiumembedded/cef/issues/3721

Additionally, the cookie directories for each service integration must be moved into the root config directory, as subdirectories in their current form (`obs_profile_cookies/<cookie_id>`) are not supported. This is intentional, and according to Marshall this was never a supported setup. cache_path *must* be a direct child of the root_cache_path. Invalid cache_path_ is silently treated as Incognito Mode and cookies are not stored.

* https://github.com/chromiumembedded/cef/issues/3930

To work around this, I've included some temporary code to translate the path received from the OBS frontend. When this PR is to be merged, I can submit a PR to obs-studio to match the intended behaviour.

Originally, I tracked down a number of crashes when loading various `chrome://` urls in the new runtime. This also exposed the fact that users would be able to access pages we didn't want them to see, such as `chrome://settings`. My original implementation therefore added a filter to block such pages from loading. Since then, CEF added its own whitelist to protect users from crashes, so my PR in its current form no longer has such protection. If it comes up, I can re-implement such protections in the future.

* The original issue https://github.com/chromiumembedded/cef/issues/3763
* The list of [allowed pages](https://github.com/chromiumembedded/cef/blob/ffa86e361d9c210136b539953663526add56191e/libcef/browser/alloy/alloy_browser_host_impl.cc#L66)
* The [check itself](https://github.com/chromiumembedded/cef/blob/ffa86e361d9c210136b539953663526add56191e/libcef/browser/alloy/alloy_browser_host_impl.cc#L564-L570)

### Motivation and Context

Keeping CEF up to date is important - for security, reliability, support, and access to modern features.

### How Has This Been Tested?

Tested on Windows 10 on CEF 147 with a `plugin_config/obs-browser` config directory from OBS 32.1 to ensure data is migrated correctly. Launched again after migration to ensure data was still accessible. `cef_binary_147.0.14+g76d2442+chromium-147.0.7727.138_windows64_minimal`

Tested again with our current CEF 127 build to ensure we can use 6533 unchanged if CEF tests themselves backfire.

Additionally tested over the past few months with the following versions as they were released. No regressions were found. Mostly my fault for taking so long.

```
cef_binary_128.0.0+g446b7d6+chromium-128.0.6613.0_windows64_minimal
cef_binary_132.3.2+g4997b2f+chromium-132.0.6834.161_windows64_minimal
cef_binary_136.1.5+g723b52b+chromium-136.0.7103.93_windows64_minimal
cef_binary_143.0.13+g30cb3bd+chromium-143.0.7499.170_windows64_minimal
cef_binary_145.0.28+g51162e8+chromium-145.0.7632.160_windows64_minimal
```

Additionally, @hoshinolina has done tremendous work testing this on Fedora, with users actively using the code in this PR since November 2025. As far as I know, users have had very few complaints.

### Types of changes

- Documentation (a change to documentation pages)

### Checklist:

### DIFF
--- a/browser-app.cpp
+++ b/browser-app.cpp
@@ -133,7 +133,7 @@ void BrowserApp::OnBeforeCommandLineProcessing(const CefString &, CefRefPtr<CefC
 
 	// See https://github.com/chromiumembedded/cef/issues/3966 for 'StorageNotificationService' requirement.
 	constexpr std::string_view kDefaultDisabledFeatures{
-		"CalculateNativeWinOcclusion,HardwareMediaKeyHandling,LiveCaption,"
+		"CalculateNativeWinOcclusion,GlobalShortcutsPortal,HardwareMediaKeyHandling,LiveCaption,"
 		"MediaRouter,StorageNotificationService,WebBluetooth,"
 		"EnableWindowsGamingInputDataFetcher"};
 
--- a/browser-app.cpp
+++ b/browser-app.cpp
@@ -48,9 +48,60 @@ CefRefPtr<CefBrowserProcessHandler> BrowserApp::GetBrowserProcessHandler()
 	return this;
 }
 
+CefRefPtr<CefClient> BrowserApp::GetDefaultClient()
+{
+	return GetDummy();
+}
+
 void BrowserApp::OnRegisterCustomSchemes(CefRawPtr<CefSchemeRegistrar> registrar)
 {
 	registrar->AddCustomScheme("http", CEF_SCHEME_OPTION_STANDARD | CEF_SCHEME_OPTION_CORS_ENABLED);
+}
+
+void BrowserApp::OnContextInitialized()
+{
+	// Without a default client, CefBrowser is unmanaged, allowing full-blown Chromium windows outside of our control.
+	// We don't actually want those, so define a dummy client which will automatically close any such windows.
+	dummy = new BrowserDummyClient();
+
+	CefRefPtr<CefRequestContext> requestContext = CefRequestContext::GetGlobalContext();
+	CefString errorMessage;
+	CefRefPtr<CefValue> optionValue = CefValue::Create();
+
+	constexpr std::array<std::string_view, 20> kBrowserFeaturesToDisable{
+		"autofill.credit_card_enabled",
+		"autofill.enabled",
+		"autofill.iban_enabled",
+		"autofill.payment_card_benefits",
+		"autofill.payment_cvc_storage",
+		"autofill.profile_enabled",
+		"autologin.enabled",
+		"browser_labs_enabled",
+		"credentials_enable_autosignin",
+		"credentials_enable_service",
+		"payments.can_make_payment_enabled",
+		"printing.enabled",
+		"search.suggest_enabled",
+		"shopping_list_enabled",
+		"side_panel.google_search_side_panel_enabled",
+		"side_search.enabled",
+		"signin.allowed",
+		"signin.allowed_on_next_startup",
+		"translate",
+		"url_keyed_anonymized_data_collection.enabled"};
+
+	constexpr std::array<std::string_view, 2> kBrowserFeaturesToEnable{"extensions.block_external_extensions",
+									   "extensions.disabled"};
+
+	optionValue->SetBool(false);
+	for (std::string_view feature : kBrowserFeaturesToDisable) {
+		requestContext->SetPreference(feature.data(), optionValue.get(), errorMessage);
+	}
+
+	optionValue->SetBool(true);
+	for (std::string_view feature : kBrowserFeaturesToEnable) {
+		requestContext->SetPreference(feature.data(), optionValue.get(), errorMessage);
+	}
 }
 
 void BrowserApp::OnBeforeChildProcessLaunch(CefRefPtr<CefCommandLine> command_line)
@@ -82,16 +133,35 @@ void BrowserApp::OnBeforeCommandLineProcessing(const CefString &, CefRefPtr<CefC
 		disableFeatures += ",EnableWindowsGamingInputDataFetcher";
 #endif
 		disableFeatures += ",WebBluetooth";
+		disableFeatures += ",MediaRouter";
+		disableFeatures += ",CalculateNativeWinOcclusion";
+		disableFeatures += ",LiveCaption";
+		// https://github.com/chromiumembedded/cef/issues/3966
+		disableFeatures += ",StorageNotificationService";
 		command_line->AppendSwitchWithValue("disable-features", disableFeatures);
 	} else {
 		command_line->AppendSwitchWithValue("disable-features", "WebBluetooth,"
 #ifdef _WIN32
 									"EnableWindowsGamingInputDataFetcher,"
 #endif
+									"MediaRouter,"
+									"CalculateNativeWinOcclusion,"
+									"LiveCaption,"
+									"StorageNotificationService,"
 									"HardwareMediaKeyHandling");
 	}
 
+	if (command_line->HasSwitch("disable-blink-features")) {
+		std::string disableBlinkFeatures = command_line->GetSwitchValue("disable-blink-features");
+		disableBlinkFeatures += ",DocumentPictureInPictureAPI";
+		command_line->AppendSwitchWithValue("disable-blink-features", disableBlinkFeatures);
+	} else {
+		command_line->AppendSwitchWithValue("disable-blink-features", "DocumentPictureInPictureAPI");
+	}
+
 	command_line->AppendSwitchWithValue("autoplay-policy", "no-user-gesture-required");
+	command_line->AppendSwitch("disable-extensions");
+	command_line->AppendSwitch("hide-crash-restore-bubble");
 #ifdef __APPLE__
 	command_line->AppendSwitch("use-mock-keychain");
 #elif !defined(_WIN32)
--- a/browser-app.cpp
+++ b/browser-app.cpp
@@ -38,6 +38,12 @@
 	}
 #endif
 
+#ifdef _WIN32
+constexpr bool kIsPlatformWindows = true;
+#else
+constexpr bool kIsPlatformWindows = false;
+#endif
+
 CefRefPtr<CefRenderProcessHandler> BrowserApp::GetRenderProcessHandler()
 {
 	return this;
@@ -125,30 +131,26 @@ void BrowserApp::OnBeforeCommandLineProcessing(const CefString &, CefRefPtr<CefC
 		}
 	}
 
+	// See https://github.com/chromiumembedded/cef/issues/3966 for 'StorageNotificationService' requirement.
+	constexpr std::string_view kDefaultDisabledFeatures{
+		"CalculateNativeWinOcclusion,HardwareMediaKeyHandling,LiveCaption,"
+		"MediaRouter,StorageNotificationService,WebBluetooth,"
+		"EnableWindowsGamingInputDataFetcher"};
+
+	std::string disableFeatures{};
+
+	if constexpr (kIsPlatformWindows) {
+		disableFeatures.append(kDefaultDisabledFeatures);
+	} else {
+		constexpr std::string_view kWindowsFeature{",EnableWindowsGamingInputDataFetcher"};
+		constexpr size_t kNonWindowsLength = kDefaultDisabledFeatures.size() - kWindowsFeature.size();
+		disableFeatures.append(kDefaultDisabledFeatures.substr(0, kNonWindowsLength));
+	}
+
 	if (command_line->HasSwitch("disable-features")) {
 		// Don't override existing, as this can break OSR
-		std::string disableFeatures = command_line->GetSwitchValue("disable-features");
-		disableFeatures += ",HardwareMediaKeyHandling";
-#ifdef _WIN32
-		disableFeatures += ",EnableWindowsGamingInputDataFetcher";
-#endif
-		disableFeatures += ",WebBluetooth";
-		disableFeatures += ",MediaRouter";
-		disableFeatures += ",CalculateNativeWinOcclusion";
-		disableFeatures += ",LiveCaption";
-		// https://github.com/chromiumembedded/cef/issues/3966
-		disableFeatures += ",StorageNotificationService";
-		command_line->AppendSwitchWithValue("disable-features", disableFeatures);
-	} else {
-		command_line->AppendSwitchWithValue("disable-features", "WebBluetooth,"
-#ifdef _WIN32
-									"EnableWindowsGamingInputDataFetcher,"
-#endif
-									"MediaRouter,"
-									"CalculateNativeWinOcclusion,"
-									"LiveCaption,"
-									"StorageNotificationService,"
-									"HardwareMediaKeyHandling");
+		disableFeatures.append(",");
+		disableFeatures.append(command_line->GetSwitchValue("disable-features"));
 	}
 
 	if (command_line->HasSwitch("disable-blink-features")) {
@@ -161,6 +163,7 @@ void BrowserApp::OnBeforeCommandLineProcessing(const CefString &, CefRefPtr<CefC
 
 	command_line->AppendSwitchWithValue("autoplay-policy", "no-user-gesture-required");
 	command_line->AppendSwitch("disable-extensions");
+	command_line->AppendSwitchWithValue("disable-features", disableFeatures);
 	command_line->AppendSwitch("hide-crash-restore-bubble");
 #ifdef __APPLE__
 	command_line->AppendSwitch("use-mock-keychain");
--- a/browser-app.hpp
+++ b/browser-app.hpp
@@ -22,6 +22,7 @@
 #include <unordered_map>
 #include <functional>
 #include "cef-headers.hpp"
+#include "browser-dummy-client.hpp"
 
 typedef std::function<void(CefRefPtr<CefBrowser>)> BrowserFunc;
 
@@ -85,6 +86,8 @@ public:
 
 	virtual CefRefPtr<CefRenderProcessHandler> GetRenderProcessHandler() override;
 	virtual CefRefPtr<CefBrowserProcessHandler> GetBrowserProcessHandler() override;
+	virtual CefRefPtr<CefClient> GetDefaultClient() override;
+	virtual void OnContextInitialized() override;
 	virtual void OnBeforeChildProcessLaunch(CefRefPtr<CefCommandLine> command_line) override;
 	virtual void OnRegisterCustomSchemes(CefRawPtr<CefSchemeRegistrar> registrar) override;
 	virtual void OnBeforeCommandLineProcessing(const CefString &process_type,
@@ -105,6 +108,10 @@ public:
 #endif
 	QTimer frameTimer;
 #endif
+
+	CefRefPtr<BrowserDummyClient> dummy = nullptr;
+
+	BrowserDummyClient *GetDummy() const { return dummy.get(); };
 
 	IMPLEMENT_REFCOUNTING(BrowserApp);
 };
--- a/browser-client.cpp
+++ b/browser-client.cpp
@@ -626,6 +626,17 @@ void BrowserClient::OnLoadEnd(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame> frame, 
 	}
 }
 
+void BrowserClient::OnLoadError(CefRefPtr<CefBrowser>, [[maybe_unused]] CefRefPtr<CefFrame> frame,
+				CefLoadHandler::ErrorCode, const CefString &, const CefString &)
+{
+#if CHROME_VERSION_BUILD > 6533
+	// CEF doesn't currently provide a way to properly disable/override the default Chrome error page.
+	// https://github.com/obsproject/obs-studio/issues/13499
+	// FIXME: https://github.com/chromiumembedded/cef/issues/3852
+	frame->LoadURL("about:blank");
+#endif
+}
+
 bool BrowserClient::OnConsoleMessage(CefRefPtr<CefBrowser>, cef_log_severity_t level, const CefString &message,
 				     const CefString &source, int line)
 {
--- a/browser-dummy-client.cpp
+++ b/browser-dummy-client.cpp
@@ -1,0 +1,87 @@
+/******************************************************************************
+    Copyright (C) 2026 by Matt Gajownik <matt@volunteer.obsproject.com>
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+******************************************************************************/
+
+#include "browser-dummy-client.hpp"
+
+CefRefPtr<CefCommandHandler> BrowserDummyClient::GetCommandHandler()
+{
+	return this;
+}
+
+CefRefPtr<CefRequestHandler> BrowserDummyClient::GetRequestHandler()
+{
+	return this;
+}
+
+CefRefPtr<CefLifeSpanHandler> BrowserDummyClient::GetLifeSpanHandler()
+{
+	return this;
+}
+
+bool BrowserDummyClient::OnChromeCommand(CefRefPtr<CefBrowser>, int, cef_window_open_disposition_t)
+{
+	return true;
+}
+
+bool BrowserDummyClient::IsChromeAppMenuItemVisible(CefRefPtr<CefBrowser>, int)
+{
+	return false;
+}
+
+bool BrowserDummyClient::IsChromeToolbarButtonVisible(cef_chrome_toolbar_button_type_t)
+{
+	return false;
+}
+
+bool BrowserDummyClient::IsChromePageActionIconVisible(cef_chrome_page_action_icon_type_t)
+{
+	return false;
+}
+
+bool BrowserDummyClient::IsChromeAppMenuItemEnabled(CefRefPtr<CefBrowser>, int)
+{
+	return false;
+}
+
+bool BrowserDummyClient::OnBeforePopup(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame>,
+#if CHROME_VERSION_BUILD >= 6834
+				       int,
+#endif
+				       const CefString &, const CefString &, cef_window_open_disposition_t, bool,
+				       const CefPopupFeatures &, CefWindowInfo &, CefRefPtr<CefClient> &,
+				       CefBrowserSettings &, CefRefPtr<CefDictionaryValue> &, bool *)
+{
+	return true;
+}
+
+void BrowserDummyClient::OnAfterCreated(CefRefPtr<CefBrowser> browser)
+{
+	if (browser && browser->GetHost()) {
+		browser->GetHost()->CloseBrowser(false);
+	}
+}
+
+bool BrowserDummyClient::OnOpenURLFromTab(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame>, const CefString &,
+					  CefRequestHandler::WindowOpenDisposition, bool)
+{
+	return true;
+}
+
+bool BrowserDummyClient::OnBeforeBrowse(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame>, CefRefPtr<CefRequest>, bool, bool)
+{
+	return true;
+}
--- a/browser-dummy-client.hpp
+++ b/browser-dummy-client.hpp
@@ -1,0 +1,64 @@
+/******************************************************************************
+    Copyright (C) 2026 by Matt Gajownik <matt@volunteer.obsproject.com>
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+******************************************************************************/
+
+#pragma once
+
+#include "cef-headers.hpp"
+
+class BrowserDummyClient : public CefClient,
+			   public CefCommandHandler,
+			   public CefRequestHandler,
+			   public CefLifeSpanHandler {
+public:
+	virtual CefRefPtr<CefCommandHandler> GetCommandHandler() override;
+	virtual CefRefPtr<CefRequestHandler> GetRequestHandler() override;
+	virtual CefRefPtr<CefLifeSpanHandler> GetLifeSpanHandler() override;
+
+	/* CefCommandHandler */
+	virtual bool OnChromeCommand(CefRefPtr<CefBrowser> browser, int command_id,
+				     cef_window_open_disposition_t disposition) override;
+	virtual bool IsChromeAppMenuItemVisible(CefRefPtr<CefBrowser> browser, int command_id) override;
+	virtual bool IsChromeToolbarButtonVisible(cef_chrome_toolbar_button_type_t button_type) override;
+
+	virtual bool IsChromePageActionIconVisible(cef_chrome_page_action_icon_type_t icon_type) override;
+
+	virtual bool IsChromeAppMenuItemEnabled(CefRefPtr<CefBrowser> browser, int command_id) override;
+
+	/* CefLifeSpanHandler */
+	virtual bool OnBeforePopup(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame,
+#if CHROME_VERSION_BUILD >= 6834
+				   int,
+#endif
+				   const CefString &target_url, const CefString &target_frame_name,
+				   cef_window_open_disposition_t target_disposition, bool user_gesture,
+				   const CefPopupFeatures &popupFeatures, CefWindowInfo &windowInfo,
+				   CefRefPtr<CefClient> &client, CefBrowserSettings &settings,
+				   CefRefPtr<CefDictionaryValue> &extra_info, bool *no_javascript_access) override;
+
+	virtual void OnAfterCreated(CefRefPtr<CefBrowser> browser) override;
+
+	/* CefRequestHandler */
+	virtual bool OnOpenURLFromTab(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame,
+				      const CefString &target_url,
+				      CefRequestHandler::WindowOpenDisposition target_disposition,
+				      bool user_gesture) override;
+
+	virtual bool OnBeforeBrowse(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame,
+				    CefRefPtr<CefRequest> request, bool user_gesture, bool is_redirect) override;
+
+	IMPLEMENT_REFCOUNTING(BrowserDummyClient);
+};
--- a/cmake/os-linux.cmake
+++ b/cmake/os-linux.cmake
@@ -14,7 +14,8 @@ add_executable(OBS::browser-helper ALIAS browser-helper)
 
 target_sources(
   browser-helper PRIVATE # cmake-format: sortable
-                         browser-app.cpp browser-app.hpp cef-headers.hpp obs-browser-page/obs-browser-page-main.cpp)
+                         browser-app.cpp browser-app.hpp browser-dummy-client.cpp browser-dummy-client.hpp
+                         cef-headers.hpp obs-browser-page/obs-browser-page-main.cpp)
 
 target_include_directories(browser-helper PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/deps"
                                                   "${CMAKE_CURRENT_SOURCE_DIR}/obs-browser-page")
--- a/cmake/os-macos.cmake
+++ b/cmake/os-macos.cmake
@@ -29,7 +29,8 @@ foreach(helper IN LISTS helper_suffixes)
 
   target_sources(
     ${target_name} PRIVATE # cmake-format: sortable
-                           browser-app.cpp browser-app.hpp cef-headers.hpp obs-browser-page/obs-browser-page-main.cpp)
+                           browser-app.cpp browser-app.hpp browser-dummy-client.cpp browser-dummy-client.hpp
+                           cef-headers.hpp obs-browser-page/obs-browser-page-main.cpp)
 
   target_compile_definitions(${target_name} PRIVATE ENABLE_BROWSER_SHARED_TEXTURE)
 
--- a/cmake/os-windows.cmake
+++ b/cmake/os-windows.cmake
@@ -13,8 +13,8 @@ add_executable(OBS::browser-helper ALIAS obs-browser-helper)
 target_sources(
   obs-browser-helper
   PRIVATE # cmake-format: sortable
-          browser-app.cpp browser-app.hpp cef-headers.hpp obs-browser-page.manifest
-          obs-browser-page/obs-browser-page-main.cpp)
+          browser-app.cpp browser-app.hpp browser-dummy-client.cpp browser-dummy-client.hpp
+          cef-headers.hpp obs-browser-page.manifest obs-browser-page/obs-browser-page-main.cpp)
 
 configure_file(cmake/windows/obs-module-helper.rc.in obs-browser-page.rc)
 target_sources(obs-browser-helper PRIVATE obs-browser-page.rc)
--- a/obs-browser-plugin.cpp
+++ b/obs-browser-plugin.cpp
@@ -293,6 +293,9 @@ static void BrowserInit(void)
 	CefString(&settings.log_file) = log_path_abs;
 	settings.windowless_rendering_enabled = true;
 	settings.no_sandbox = true;
+#if CHROME_VERSION_BUILD > 6533 && CHROME_VERSION_BUILD <= 6613
+	settings.chrome_runtime = true;
+#endif
 
 	uint32_t obs_ver = obs_get_version();
 	uint32_t obs_maj = obs_ver >> 24;
@@ -341,6 +344,9 @@ static void BrowserInit(void)
 	settings.persist_user_preferences = 1;
 #endif
 	CefString(&settings.cache_path) = conf_path_abs;
+#if CHROME_VERSION_BUILD > 6533
+	CefString(&settings.root_cache_path) = conf_path_abs;
+#endif
 #if !defined(__APPLE__) || defined(ENABLE_BROWSER_LEGACY)
 	char *abs_path = os_get_abs_path_ptr(path.c_str());
 	CefString(&settings.browser_subprocess_path) = abs_path;
@@ -714,8 +720,122 @@ static void check_hwaccel_support(void)
 #endif
 #endif
 
+#if CHROME_VERSION_BUILD > 6533
+static void MigrateProfileConfigDir(std::string oldDir, std::string newDir, bool makeDirs)
+{
+	namespace fs = std::filesystem;
+
+	BPtr<char> configPath = obs_module_config_path(".");
+	fs::path rootPath{configPath.Get()};
+
+	fs::path oldPath = rootPath / fs::path(oldDir);
+	fs::path newPath = rootPath / fs::path(newDir);
+
+	if (!fs::exists(oldPath)) {
+		return;
+	}
+
+	try {
+		if (makeDirs && !fs::exists(newPath)) {
+			fs::create_directories(newPath);
+		}
+		const auto copyOptions = fs::copy_options::recursive;
+		fs::copy(oldPath, newPath, copyOptions);
+	} catch (const fs::filesystem_error &error) {
+		blog(LOG_WARNING, "[obs-browser]: Error migrating cookies for '%s': %s", oldPath.c_str(), error.what());
+	}
+}
+
+static void MigrateDefaultProfileSubdirs(std::string parent, bool makeDirs)
+{
+	constexpr std::array<std::string_view, 3> kMigrationDirectories{
+		"Local Storage",
+		"Session Storage",
+		"Network",
+	};
+	constexpr std::string_view kDefaultDirectoryPrefix{"Default/"};
+
+	for (std::string_view directory : kMigrationDirectories) {
+		std::string oldDirectory{parent};
+		oldDirectory.append(directory);
+		std::string newDirectory{parent};
+		newDirectory.append(kDefaultDirectoryPrefix);
+		newDirectory.append(directory);
+
+		MigrateProfileConfigDir(oldDirectory, newDirectory, makeDirs);
+	}
+}
+
+static void MigrateToChromeRuntime()
+{
+	namespace fs = std::filesystem;
+	BPtr<char> defaultConfigPath = obs_module_config_path("Default");
+	fs::path defaultPath{defaultConfigPath.Get()};
+	// User has already previously launched with Chrome Runtime
+	if (fs::exists(defaultPath)) {
+		blog(LOG_INFO, "[obs-browser]: Chrome Runtime Default profile detected. No migration necessary.");
+		return;
+	}
+
+	BPtr<char> localPrefsConfigPath = obs_module_config_path("LocalPrefs.json");
+	fs::path localPrefs{localPrefsConfigPath.Get()};
+	// User has never launched old OBS Browser before
+	if (!fs::exists(localPrefs)) {
+		blog(LOG_DEBUG, "[obs-browser]: Alloy Runtime preferences not found. No migration necessary.");
+		return;
+	}
+
+	blog(LOG_INFO, "[obs-browser]: Migrating files to Chrome Runtime...");
+
+	try {
+		fs::create_directories(defaultPath);
+	} catch (const fs::filesystem_error &error) {
+		blog(LOG_WARNING, "[obs-browser]: Error creating Default profile: %s", error.what());
+	}
+
+	MigrateDefaultProfileSubdirs("", false);
+
+	BPtr<char> newLocalPrefsConfigPath = obs_module_config_path("Local State");
+	fs::path newLocalPrefs{newLocalPrefsConfigPath.Get()};
+	try {
+		fs::copy(localPrefs, newLocalPrefs);
+	} catch (const fs::filesystem_error &error) {
+		blog(LOG_WARNING, "[obs-browser]: Error migrating preferences: %s", error.what());
+	}
+
+	constexpr std::string_view prefix = "obs_profile_cookies";
+
+	BPtr<char> serviceProfilesPath = obs_module_config_path(prefix.data());
+	fs::path serviceProfiles{serviceProfilesPath.Get()};
+	if (fs::exists(serviceProfiles)) {
+		// User has service integration cookies that also must be copied
+		for (auto const &dir_entry : std::filesystem::directory_iterator{serviceProfiles}) {
+			if (!dir_entry.is_directory()) {
+				continue;
+			}
+			std::string name{dir_entry.path().filename().string()};
+			std::string oldDir{prefix};
+			oldDir.append("/");
+			oldDir.append(name);
+
+			std::string newDir{prefix};
+			newDir.append("_");
+			newDir.append(name);
+			MigrateProfileConfigDir(oldDir, newDir, true);
+
+			blog(LOG_INFO, "[obs-browser]: Migrated cookies for %s.", name.c_str());
+		}
+	}
+
+	blog(LOG_INFO, "[obs-browser]: Migration of browser cookies and session completed.");
+}
+#endif
+
 bool obs_module_load(void)
 {
+#if CHROME_VERSION_BUILD > 6533
+	MigrateToChromeRuntime();
+#endif
 #ifdef ENABLE_BROWSER_QT_LOOP
 	qRegisterMetaType<MessageTask>("MessageTask");
 #endif
--- a/panel/browser-panel-client.cpp
+++ b/panel/browser-panel-client.cpp
@@ -38,6 +38,13 @@ CefRefPtr<CefDisplayHandler> QCefBrowserClient::GetDisplayHandler()
 	return this;
 }
 
+#if CHROME_VERSION_BUILD >= 6533
+CefRefPtr<CefCommandHandler> QCefBrowserClient::GetCommandHandler()
+{
+	return this;
+}
+#endif
+
 CefRefPtr<CefRequestHandler> QCefBrowserClient::GetRequestHandler()
 {
 	return this;
@@ -68,10 +75,18 @@ CefRefPtr<CefJSDialogHandler> QCefBrowserClient::GetJSDialogHandler()
 	return this;
 }
 
+/* CefCommandHandler */
+#if CHROME_VERSION_BUILD >= 6533
+bool QCefBrowserClient::OnChromeCommand(CefRefPtr<CefBrowser>, int, cef_window_open_disposition_t)
+{
+	return true;
+}
+#endif
+
 /* CefDisplayHandler */
 void QCefBrowserClient::OnTitleChange(CefRefPtr<CefBrowser> browser, const CefString &title)
 {
-	if (widget && widget->cefBrowser->IsSame(browser)) {
+	if (widget && widget->cefBrowser && widget->cefBrowser->IsSame(browser)) {
 		std::string str_title = title;
 		QString qt_title = QString::fromUtf8(str_title.c_str());
 		QMetaObject::invokeMethod(widget, "titleChanged", Q_ARG(QString, qt_title));
--- a/panel/browser-panel.cpp
+++ b/panel/browser-panel.cpp
@@ -368,6 +368,10 @@ void QCefWidgetInternal::resizeEvent(QResizeEvent *event)
 
 void QCefWidgetInternal::Resize()
 {
+	if (os_event_try(cef_started_event) != 0) {
+		return;
+	}
+
 	QSize size = this->size() * devicePixelRatioF();
 
 	bool success = QueueCEFTask([this, size]() {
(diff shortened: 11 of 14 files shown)


hoshinolina:
For reference, the only issues we've run into in Fedora as far as I remember have been:

* Unrelated bugs in OBS (that affected CEF/obs-browser for computers-are-hard-tm reasons)
* #493 & #491 (merged)
* #519 (this has always been broken but newer CEF makes it assert way more often when it fails the ABI version check, you probably want to get that one sorted out / added as a dependency to this)
* Unrelated bugs in CEF (since fixed)
* Binary compatibility issues: Newer versions of CEF have a stable ABI that allows decoupling CEF updates from OBS updates. Fedora takes advantage of this to ship CEF as a generic package (possibly usable by other consumers) and keep it as up to date as possible (in fact, Fedora decouples CEF and Chromium updates too, sometimes shipping Chromium security fixes within the CEF package before upstream CEF itself). This mechanism has had some teething issues with unexpected compat breaks. As long as OBS continues to ship updates in lockstep with CEF and against the `EXPERIMENTAL` API version (which is the default), this should not affect it (in all cases, rebuilding OBS against the current version/API of CEF would have fixed/worked around the ABI break we saw).
* The whole no-docks-in-Wayland mess, which has nothing to do with this

So overall, nothing really relevant to this PR. That said, I should point out that Fedora never shipped nor tested the profile migration code. While Fedora did briefly ship the old obs-cef at one point in the past, that was abandoned as unmaintainable (making it build with newer compilers/libs became a nightmare), so there was a period where OBS had no CEF support in Fedora before this was reintroduced, and therefore no good reason to worry about profile migrations (it would only benefit people who had used the old version, to revive their old dormant profiles, which we figured wasn't really worth it).

If anyone wants to build this PR in Fedora against system CEF, the only required changes should be:

1. Install `cef-devel`
2. `rm cmake/finders/FindCEF.cmake` (cef-devel provides this systemwide, interface-compatible with the OBS version)
3. Either run with `LD_LIBRARY_PATH=/usr/lib64/cef` or change the rpath setting:
    `sed -e 's,INSTALL_RPATH ".*",INSTALL_RPATH "/usr/lib64/cef/",' -i plugins/obs-browser/cmake/os-linux.cmake`
4. `-DENABLE_BROWSER=ON` and optionally, `-DCEF_API_VERSION=xxxxx` if you want to build against a specific versioned API/ABI (Fedora arbitrarily picks 13700 here, but I don't think CEF has actually changed any APIs in a source-incompatible way with OBS since 13300, the very first version, so you can pick whatever right now).

Note that unlike the standard OBS build, this setup embeds the `libcef_dll_wrapper` build into the OBS build, so you will see that being built alongside the rest of obs-browser (the wrapper source code is shipped in `cef-devel`). All that is taken care of by the system FindCEF.cmake and the way the CEF package itself is patched/configured, and in fact it turned out to be quite easy to do and it works very well. We settled on this approach after trying to pre-build the wrapper in `cef-devel`, because we realized that it was impractical to build a separate wrapper for every possible version and additionally, consumers may want to build the wrapper with their own choice of compile-time options making "universal" wrapper builds not really a good idea (upstream CEF confirmed this too). Relevant links: [FindCEF.cmake](https://src.fedoraproject.org/rpms/cef/blob/rawhide/f/FindCEF.cmake) [cef.spec](https://src.fedoraproject.org/rpms/cef/blob/rawhide/f/cef.spec) (search for libcef_dll for the few relevant lines). I would personally encourage OBS to consider exploring this approach instead of a separate libcef_dll_wrapper build for official binaries/CI, but that's just me (and it's not a blocker for this PR in any way) \^\^

panel/browser-panel.cpp
```diff
@@ -552,8 +552,18 @@ QCefWidget *QCefInternal::create_widget(QWidget *parent, const std::string &url,
 
 QCefCookieManager *QCefInternal::create_cookie_manager(const std::string &storage_path, bool persist_session_cookies)
 {
+	std::string path = storage_path;
+#if CHROME_VERSION_BUILD > 6533
+	// This code may be temporary, ideally the change should be in OBS Studio itself,
```
PatTheMav: Create a task issue and add a `TODO` comment with a link to that issue here (outlining the issue and what needs to be changed in that issue).

obs-browser-plugin.cpp
```diff
@@ -293,6 +293,9 @@ static void BrowserInit(void)
 	CefString(&settings.log_file) = log_path_abs;
 	settings.windowless_rendering_enabled = true;
 	settings.no_sandbox = true;
+#if CHROME_VERSION_BUILD > 6533 && CHROME_VERSION_BUILD < 7103
```
PatTheMav: What's the reason to guard this against versions 7103 and up?

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,97 @@ static void check_hwaccel_support(void)
...
+
+	for (std::string oldDir : dirs) {
+		std::string newDir("Default/" + oldDir);
+		MigrateProfileConfigDir(parent + oldDir, parent + newDir, makeDirs);
+	}
+}
+
+static void MigrateToChromeRuntime()
+{
+	BPtr<char> defaultProfile = obs_module_config_path("Default");
```
PatTheMav: This is entirely new C++ code, so I'd prefer if we'd use standard C++ library types and functions if available (e.g. `std::string`, `std::string_view`, `std::filesystem`, etc.) rather than custom OBS types and APIs and treating it as "C with some OOP".

(And even though `std` is included by default, this would be a violation of our C++ code style standards, so use `std::string` explicitly so this code does not need to be touched when `using namespace std` hopefully is excised at a later point).

Also be mindful of any `libobs` API that potentially returns `nullptr` instead of an actual `char *` and guard against that.

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,97 @@ static void check_hwaccel_support(void)
...
+}
+
+static void MigrateDefaultProfileSubdirs(std::string parent, bool makeDirs)
+{
+	std::string dirs[3] = {"Local Storage", "Session Storage", "Network"};
+
+	for (std::string oldDir : dirs) {
+		std::string newDir("Default/" + oldDir);
+		MigrateProfileConfigDir(parent + oldDir, parent + newDir, makeDirs);
+	}
```
PatTheMav: Use `constexpr` for the constants:
```C++
constexpr std::array<std::string_view, 3> migrationDirectories {
    "Local Storage",
    "Session Storage",
    "Network",
};

constexpr std::string_view defaultDirectoryPrefix {"Default/"};
```

Which enables:

```suggestion
	for (std::string_view directory, migrationDirectories) {
	    std::string oldDirectory {parent};
	    oldDirectory.append(directory);

	    std::string newDirectory {parent};
	    newDirectory.append(defaultDirectoryPrefix);
	    newDirectory.append(directory)

		MigrateProfileConfigDir(oldDirectory, newDirectory, makeDirs);
	}
```

I don't know if I like the design with this function and `MigrateProfileConfigDir` rather than one single function. I'm kinda fine with using string concatenation at first rather than using `std::filesystem::path` objects as any string token would have to be converted using `u8path` first due to Windows using UTF-16 for paths internally.

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,97 @@ static void check_hwaccel_support(void)
...
+
+	const std::filesystem::path old_path = std::filesystem::u8path(path_abs.Get());
+	if (!std::filesystem::exists(old_path)) {
+		return;
+	}
+
+	try {
+		const auto copyOptions = std::filesystem::copy_options::recursive;
+		std::filesystem::copy(old_path, std::string(new_path_abs.Get()), copyOptions);
```
PatTheMav: My hunch is that this might break in unforeseen ways as you pass in a `std::string` with UTF-8 bytes into a function that expects UTF-16 bytes instead.

browser-client.cpp
```diff
@@ -626,6 +626,17 @@ void BrowserClient::OnLoadEnd(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame> frame,
 	}
 }
 
+void BrowserClient::OnLoadError(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame> frame, CefLoadHandler::ErrorCode,
+				const CefString &, const CefString &)
+{
+#if CHROME_VERSION_BUILD > 6533
+	// https://github.com/chromiumembedded/cef/issues/3852
```
PatTheMav: Some high-level explanation of the issue in addition to the issue link would be helpful. Also a `FIXME` or `TODO` lead-in would be great as most IDEs and syntax-highlighters detect those.

browser-app.cpp
```diff
@@ -48,11 +48,53 @@ CefRefPtr<CefBrowserProcessHandler> BrowserApp::GetBrowserProcessHandler()
...
+}
+
 void BrowserApp::OnRegisterCustomSchemes(CefRawPtr<CefSchemeRegistrar> registrar)
 {
 	registrar->AddCustomScheme("http", CEF_SCHEME_OPTION_STANDARD | CEF_SCHEME_OPTION_CORS_ENABLED);
 }
 
+void BrowserApp::OnContextInitialized()
+{
+	dummy = new BrowserDummyClient();
```
PatTheMav, line 71: What's the purpose of the "dummy" client? Some comment explaining the technical decision making process behind it would be helpful.

browser-app.cpp
```diff
@@ -48,11 +48,53 @@ CefRefPtr<CefBrowserProcessHandler> BrowserApp::GetBrowserProcessHandler()
...
+
+	CefRefPtr<CefRequestContext> rc = CefRequestContext::GetGlobalContext();
+	CefString err;
+	CefRefPtr<CefValue> off = CefValue::Create();
+	off->SetBool(false);
+	std::string opts[20] = {"autofill.credit_card_enabled",
```
PatTheMav: Use `constexpr` for the constants:

```C++
constexpr std::array<std::string_view, 20> browserFeaturesToDisable {
[...]
};

constexpr std::array<std::string_view, 2> browserFeaturesToEnable {
[...]
};
```

```suggestion
	CefRefPtr<CefRequestContext> requestContext = CefRequestContext::GetGlobalContext();
	CefString errorMessage;
	CefRefPtr<CefValue> optionValue = CefValue::Create();
	optionValue->SetBool(false);

for (std::string_view feature : browserFeaturesToDisable) {
    requestContext->SetPreference(feature.data(), optionValue.get(), errorMessage);
}

optionValue->SetBool(true);
for (std::string_view feature: browserFeatuesToEnable) {
    requestContext->SetPreference(feature.data(), optionValue.get(), errorMessage);
}    
```

I also wonder if this should maybe be wrapped in a lambda for each value, because any error while disabling this feature should be considered a critical failure (it would leave the browser with partially enabled Chromium runtime features which in turn could lead to unexpected behaviour and related support burden).

Phr3d13:
how would I test this on arch? (sorry, really new to arch at the moment)

browser-app.cpp
PatTheMav: Forgot to mention - as per our code style guidelines those `constexpr` would also have to use the `k` prefix (`kBrowserFeaturesToDisable`).

tippfehlr:
It’s very nice to see this :tada:

Also – @hoshinolina: thanks for the CEF package in Fedora, it was very helpful when creating the Arch package ^^

@Phr3d13: I pushed `obs-studio 32.1.2-5` and `obs-studio-plugin-browser 32.1.2-5` with these changes to the [extra-testing] repository. Look at the Arch Wiki on how to [use the testing repositories](https://wiki.archlinux.org/title/Official_repositories#Testing_repositories). Though be warned, that will install all testing packages, not just obs.

panel/browser-panel.cpp
WizardCM: Can do.

browser-app.cpp
WizardCM: I documented it in the commit description and the PR description, but I'll add the note in code as well.

browser-app.cpp
WizardCM: Easy enough. Can do.

browser-client.cpp
WizardCM: Can do.

obs-browser-plugin.cpp
WizardCM: During the transition period, CEF shipped with a toggle. After CEF 128 (6613), it was no longer needed because the Alloy Runtime was entirely removed - therefore the toggle would error. This ensures compatibility with intermediary versions, though if we wanted to we could lock it purely to 6613. I left it in because it was, on occasion, useful in 6533 (127, what we currently ship) to test with.

obs-browser-plugin.cpp
WizardCM: Hmm, probably. Do we have documentation on how best to transfer from `obs_module_config_path` to `std::filesystem` cleanly? I have tried and failed to find out. I'd guess the same way I did `old_path` a few lines earlier? Just seems a bit noisy going obs_module_config_path -> os_get_abs_path_ptr -> `std::filesystem::u8path`, unless the `os_get_abs_path_ptr` can be skipped?

obs-browser-plugin.cpp
WizardCM: Same question as above for how best to transition from `obs_module_config_path` here, as I don't think we have any existing examples in the OBS codebase.

Unfortunately, `MigrateProfileConfigDir` is used **both** for migrating the Default profile, and later to migrate individual Profile cookie directories, so it has to be reusable.

Good point on catching `nullptr`s. I'll revisit.

obs-browser-plugin.cpp
PatTheMav: If it leads to an error starting with 6613, shouldn't it then be limited to versions before 6613 and not potentially lead to errors for all versions up to 7103?

obs-browser-plugin.cpp
PatTheMav: Without consulting a compiler, I'd probably do:

```C++
using fs = std::filesystem;
// This crashes if libobs returns a nullptr here, which could be seen as a 
// critical failure that should never happen.
fs::path rootPath = fs:u8path(os_module_config_path());
fs::path oldPath = rootPath / fs:u8path(oldDir);
fs::path newPath = rootPath / fs:u8path(newDir);

if (makeDirs && !fs::exists(newPath)) {
    fs::createDirectory(newPath);
}

[..]
```
On macOS `std::filesystem` seems to handle absolute paths with relative path tokens just fine (so something like `/Path/../SomePath` works), otherwise `std::filesystem::canonical` can be used to resolve such a path completely, which you'd probably wrap the assignment to `oldPath` and `newPath` in - but I'm not sure that's necessary.

obs-browser-plugin.cpp
PatTheMav: I kinda agree with @notr1ch that we don't need to use `nullptr` checks if the occurrence of it would be a truly exceptional situation (as in: this should never happen and if it does, some developer messed up).

This would be in opposition to an API that could very well return a `nullptr` and that's part of its "normal operation".

The last consideration is whether even the exceptional situation is indeed "recoverable" or doing so would allow OBS to limp on as a program in a weird, undefined, state, which could lead to data corruption or data loss as a result.

Just crashing (to avoid that) is incredibly disruptive, but preferable to e.g. potentially overwriting a scene collection with garbage data.

obs-browser-plugin.cpp
WizardCM: Yeah, I re-checked the CEF docs and it's gone as early as 6613, so I've updated the check.

obs-browser-plugin.cpp
WizardCM: Done. Let me know if there are additional concerns outside of the lack of u8path handling due to C++20 deprecation warnings.

obs-browser-plugin.cpp
WizardCM: Yep, that code worked pretty much out of the box - feel free to look over my changes. I have verified `std::filesystem::canonical` was not necessary (though it could be used to make error logging a bit clearer).

Changes made, outside of properly handing u8paths for now.

browser-client.cpp
WizardCM: Done.

browser-app.cpp
WizardCM: Note added.

browser-app.cpp
WizardCM: Changes made outside of swapping to a lambda. I'll need to think on it some more, as majority of the features being disabled are mostly to avoid quirks. At the very least we could log the ones that fail to apply so we know to fix them. Happy to revisit.

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,112 @@ static void check_hwaccel_support(void)
 #endif
 #endif
 
+#if CHROME_VERSION_BUILD > 6533
+static void MigrateProfileConfigDir(std::string oldDir, std::string newDir, bool makeDirs)
+{
+	namespace fs = std::filesystem;
+	fs::path rootPath(obs_module_config_path("."));
```
tytan652: Leaks the C string

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,112 @@ static void check_hwaccel_support(void)
...
+		newDirectory.append(directory);
+
+		MigrateProfileConfigDir(oldDirectory, newDirectory, makeDirs);
+	}
+}
+
+static void MigrateToChromeRuntime()
+{
+	namespace fs = std::filesystem;
+	fs::path defaultPath(obs_module_config_path("Default"));
```
tytan652: Leaks the C string

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,112 @@ static void check_hwaccel_support(void)
...
+	// User has already previously launched with Chrome Runtime
+	if (fs::exists(defaultPath)) {
+		blog(LOG_INFO, "[obs-browser]: Chrome Runtime Default profile detected. No migration necessary.");
+		return;
+	}
+
+	fs::path localPrefs(obs_module_config_path("LocalPrefs.json"));
```
tytan652: Leaks the C string

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,112 @@ static void check_hwaccel_support(void)
...
+
+	try {
+		fs::create_directories(defaultPath);
+	} catch (const fs::filesystem_error &error) {
+		blog(LOG_WARNING, "[obs-browser]: Error creating Default profile: %s", error.what());
+	}
+
+	MigrateDefaultProfileSubdirs("", false);
+
+	fs::path newLocalPrefs(obs_module_config_path("Local State"));
```
tytan652: Leaks the C string

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,112 @@ static void check_hwaccel_support(void)
...
+	try {
+		fs::copy(localPrefs, newLocalPrefs);
+	} catch (const fs::filesystem_error &error) {
+		blog(LOG_WARNING, "[obs-browser]: Error migrating preferences: %s", error.what());
+	}
+
+	std::string prefix = "obs_profile_cookies";
+
+	fs::path serviceProfiles(obs_module_config_path(prefix.c_str()));
```
tytan652: Leaks the C string

tytan652 (changes requested):
C strings leaks…

panel/browser-panel-client.cpp
```diff
@@ -68,10 +75,19 @@ CefRefPtr<CefJSDialogHandler> QCefBrowserClient::GetJSDialogHandler()
 	return this;
 }
 
+/* CefCommandHandler */
+#if CHROME_VERSION_BUILD >= 6533
+bool QCefBrowserClient::OnChromeCommand(CefRefPtr<CefBrowser> browser, int command_id,
+					cef_window_open_disposition_t disposition)
```
tytan652: Unused params
```suggestion
bool QCefBrowserClient::OnChromeCommand(CefRefPtr<CefBrowser> , int, cef_window_open_disposition_t)
```

browser-app.hpp
```diff
@@ -106,5 +109,9 @@ class BrowserApp : public CefApp, public CefRenderProcessHandler, public CefBrow
 	QTimer frameTimer;
 #endif
 
+	CefRefPtr<BrowserDummyClient> dummy = nullptr;
+
+	BrowserDummyClient *const GetDummy() { return dummy.get(); };
```
tytan652:
```
error: type qualifiers ignored on function return type [-Werror=ignored-qualifiers]
  114 |         BrowserDummyClient *const GetDummy() { return dummy.get(); };
      |         ^~~~~~~~~~~~~~~~~~
```

tytan652 (changes requested):
Issue met while building on Arch and Flatpak.

obs-browser-plugin.cpp
WizardCM: Fixed using my original solution. If Pat knows of a cleaner way, I'll make the additional adjustment.

obs-browser-plugin.cpp
WizardCM: Fixed.

obs-browser-plugin.cpp
WizardCM: Fixed.

panel/browser-panel-client.cpp
WizardCM: Oops, not sure how I missed that one. Fixed.

browser-app.hpp
WizardCM: I don't fully understand the error message itself, but I think I've fixed it. Please confirm.

tytan652: It works

tytan652: > I don't fully understand the error message itself

The type was `BrowserDummyClient *const` but the function return `BrowserDummyClient *` so mismatch, the `const` qualifier was assigned to the type not to the function.

browser-app.cpp
PatTheMav: Requires use of C++-style comments and not C-style comments, per code style guidelines.

browser-app.cpp
```diff
@@ -48,11 +48,64 @@ CefRefPtr<CefBrowserProcessHandler> BrowserApp::GetBrowserProcessHandler()
...
+		"search.suggest_enabled",
+		"shopping_list_enabled",
+		"side_panel.google_search_side_panel_enabled",
+		"side_search.enabled",
+		"signin.allowed",
+		"signin.allowed_on_next_startup",
+		"translate",
+		"url_keyed_anonymized_data_collection.enabled"};
+
+	constexpr std::array<std::string_view, 2> kBrowserFeaturesToEnable{"extensions.block_external_extensions",
```
PatTheMav, line 99: I guess this is `clang-format`'s doing? 🙈

browser-client.cpp
```diff
@@ -626,6 +626,18 @@ void BrowserClient::OnLoadEnd(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame> frame,
 	}
 }
 
+void BrowserClient::OnLoadError(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame> frame, CefLoadHandler::ErrorCode,
+				const CefString &, const CefString &)
+{
+#if CHROME_VERSION_BUILD > 6533
+	// CEF doesn't currently provide a way to properly disable/override the default Chrome error page
+	// FIXME https://github.com/chromiumembedded/cef/issues/3852
```
PatTheMav:
```suggestion
	// FIXME: https://github.com/chromiumembedded/cef/issues/3852
```

panel/browser-panel.cpp
```diff
@@ -552,8 +552,17 @@ QCefWidget *QCefInternal::create_widget(QWidget *parent, const std::string &url,
 
 QCefCookieManager *QCefInternal::create_cookie_manager(const std::string &storage_path, bool persist_session_cookies)
 {
+	std::string path = storage_path;
+#if CHROME_VERSION_BUILD > 6533
+	// TODO Update obs-studio to use the new path structure due to subdirectories not being supported by CEF
```
PatTheMav:
```suggestion
	// TODO: Update obs-studio to use the new path structure due to subdirectories not being supported by CEF
```

Is also missing a link to the task issue on the main repo.

Applies below as well.

browser-client.cpp
```diff
@@ -626,6 +626,18 @@ void BrowserClient::OnLoadEnd(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame> frame,
...
+	frame->LoadURL("about:blank");
+#else
+	UNUSED_PARAMETER(frame);
```
PatTheMav: Wouldn't be sufficient to decorate the `frame` argument with `[[maybe_unused]]`?

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,118 @@ static void check_hwaccel_support(void)
 #endif
 #endif
 
+#if CHROME_VERSION_BUILD > 6533
+static void MigrateProfileConfigDir(std::string oldDir, std::string newDir, bool makeDirs)
+{
+	namespace fs = std::filesystem;
+
+	BPtr<char> configPath = obs_module_config_path(".");
+	fs::path rootPath(configPath.Get());
```
PatTheMav:
```suggestion
	fs::path rootPath{configPath.Get()};
```

Prefer direct-list-initialization to catch possible narrowing conversions.

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,118 @@ static void check_hwaccel_support(void)
...
+
+		MigrateProfileConfigDir(oldDirectory, newDirectory, makeDirs);
+	}
+}
+
+static void MigrateToChromeRuntime()
+{
+	namespace fs = std::filesystem;
+	BPtr<char> defaultConfigPath = obs_module_config_path("Default");
+	fs::path defaultPath(defaultConfigPath.Get());
```
PatTheMav:
```suggestion
	fs::path defaultPath{defaultConfigPath.Get()};
```
See above.

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,118 @@ static void check_hwaccel_support(void)
...
+	// User has already previously launched with Chrome Runtime
+	if (fs::exists(defaultPath)) {
+		blog(LOG_INFO, "[obs-browser]: Chrome Runtime Default profile detected. No migration necessary.");
+		return;
+	}
+
+	BPtr<char> localPrefsConfigPath = obs_module_config_path("LocalPrefs.json");
+	fs::path localPrefs(localPrefsConfigPath.Get());
```
PatTheMav:
```suggestion
	fs::path localPrefs{localPrefsConfigPath.Get()};
```

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,118 @@ static void check_hwaccel_support(void)
...
+	try {
+		fs::create_directories(defaultPath);
+	} catch (const fs::filesystem_error &error) {
+		blog(LOG_WARNING, "[obs-browser]: Error creating Default profile: %s", error.what());
+	}
+
+	MigrateDefaultProfileSubdirs("", false);
+
+	BPtr<char> newLocalPrefsConfigPath = obs_module_config_path("Local State");
+	fs::path newLocalPrefs(newLocalPrefsConfigPath.Get());
```
PatTheMav:
```suggestion
	fs::path newLocalPrefs{newLocalPrefsConfigPath.Get()};
```

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,118 @@ static void check_hwaccel_support(void)
...
+	try {
+		fs::copy(localPrefs, newLocalPrefs);
+	} catch (const fs::filesystem_error &error) {
+		blog(LOG_WARNING, "[obs-browser]: Error migrating preferences: %s", error.what());
+	}
+
+	std::string prefix = "obs_profile_cookies";
```
PatTheMav: Replace with `constexpr std::string_view` constant.

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,118 @@ static void check_hwaccel_support(void)
...
+
+	BPtr<char> serviceProfilesPath = obs_module_config_path(prefix.c_str());
+	fs::path serviceProfiles(serviceProfilesPath.Get());
```
PatTheMav:
```suggestion
	fs::path serviceProfiles{serviceProfilesPath.Get()};
```

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,118 @@ static void check_hwaccel_support(void)
...
+	if (fs::exists(serviceProfiles)) {
+		// User has service integration cookies that also must be copied
+		for (auto const &dir_entry : std::filesystem::directory_iterator{serviceProfiles}) {
+			std::string name(dir_entry.path().filename().string());
```
PatTheMav:
```suggestion
			std::string name{dir_entry.path().filename().u8string()};
```

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,118 @@ static void check_hwaccel_support(void)
...
+			if (!dir_entry.is_directory()) {
+				continue;
+			}
+			std::string oldDir(prefix + "/" + name);
+			std::string newDir(prefix + "_" + name);
+			MigrateProfileConfigDir(oldDir, newDir, true);
+			// Subdirectories do not need to be migrated because it is not the Default profile
```
PatTheMav: What does this comment relate to? The call to `MigrateProfileConfigDir`?

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,118 @@ static void check_hwaccel_support(void)
...
+
+	BPtr<char> serviceProfilesPath = obs_module_config_path(prefix.c_str());
+	fs::path serviceProfiles(serviceProfilesPath.Get());
+	if (fs::exists(serviceProfiles)) {
+		// User has service integration cookies that also must be copied
+		for (auto const &dir_entry : std::filesystem::directory_iterator{serviceProfiles}) {
+			std::string name(dir_entry.path().filename().string());
+			if (!dir_entry.is_directory()) {
+				continue;
+			}
```
PatTheMav, line 815: Wouldn't it make more sense to move this check to before extracting the string (which will create temporary path objects on the way there) and bail out early?

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,118 @@ static void check_hwaccel_support(void)
...
+			std::string oldDir(prefix + "/" + name);
+			std::string newDir(prefix + "_" + name);
```
PatTheMav:
```suggestion
			std::string oldDir{prefix};
			oldDir.append("/");
			oldDir.append(name);

			std::string newDir{prefix};
			newDir.append("_");
			newDir.append(name);
```
Makes it easier to debug the code and keeps the amount of temporaries low.

The existing implementation is equivalent to:
```C++
std::string temp1 = prefix;
std::string temp2 = temp1.append("/");
std::string oldDir = temp2.append(name);
```
Which potentially involves a whole lot of copies after reallocations, and at least according to compiler explorer the resulting code uses about twice as many instructions even with `-O3` optimisation.

browser-client.cpp
WizardCM: It would be the first instance in the entire OBS codebase, but I'm not against trying something new!

browser-app.cpp
WizardCM: It is, yes. :)

obs-browser-plugin.cpp
WizardCM: It was a leftover comment with the intent of describing why this particular path doesn't manually copy Local Storage, Session Storage or Network, but thinking on it now it's unnecessary.

obs-browser-plugin.cpp
WizardCM: >  no instance of constructor matches the argument list
            argument types are: (std::basic_string<char8_t, std::char_traits<char8_t>, std::allocator<char8_t>>)

obs-browser-plugin.cpp
WizardCM: Done.

obs-browser-plugin.cpp
WizardCM: Done.

obs-browser-plugin.cpp
WizardCM: Done.

obs-browser-plugin.cpp
WizardCM: Done. Switched to using `.data()` for the libobs call.

obs-browser-plugin.cpp
WizardCM: Done.

obs-browser-plugin.cpp
WizardCM: You're right. Fixed.

obs-browser-plugin.cpp
WizardCM: Unfortunate, I love a one-liner. Fixed.

obs-browser-plugin.cpp
WizardCM: Done.

browser-app.cpp
WizardCM: Oops, my JavaScript Developer is showing. Fixed.

browser-client.cpp
WizardCM: Seems to build fine on both Windows and Linux, therefore marking as resolved.

panel/browser-panel.cpp
WizardCM: Took me a bit to figure out how to create a Task. Done.

browser-client.cpp
WizardCM: Done, and created Task.

panel/browser-panel.cpp
```diff
@@ -562,7 +572,16 @@ QCefCookieManager *QCefInternal::create_cookie_manager(const std::string &storag
 
 BPtr<char> QCefInternal::get_cookie_path(const std::string &storage_path)
 {
-	BPtr<char> rpath = obs_module_config_path(storage_path.c_str());
+	std::string path = storage_path;
+#if CHROME_VERSION_BUILD > 6533
+	// TODO Update obs-studio to use the new path structure due to subdirectories not being supported by CEF
```
PatTheMav:
```suggestion
	// TODO: Update obs-studio to use the new path structure due to subdirectories not being supported by CEF
	// https://github.com/obsproject/obs-studio/issues/13498
```

PatTheMav (commented):
Looks fine otherwise. Haven't given this a test on my machine yet though, will do so soon.

panel/browser-panel.cpp
WizardCM: Fixed

WizardCM:
I've rebased and pushed the latest review fixes, with a couple minor additions

1. An `os_event_try(cef_started_event)` check to `QCefWidgetInternal::Resize()` to fix a crash when the user has no browser sources but does have browser docks. This is initiated by the `restoreDockState` function before CEF has initialised. Docks do correctly resize once they've loaded.
    * I could instead put this in `bool QueueCEFTask` for ultimate protection. Open to feedback on this.
3. Disabling of `StorageNotificationService` just in case
    * https://github.com/chromiumembedded/cef/issues/3966

Additionally, after some testing from Warchamp, we have confirmed the two following regressions

1. The remote debugging hosted webpage doesn't load. Current running theory is the `shell_devtools_discovery_page.html` resource is not being shipped in the `.pak` file by CEF, so the website returns an empty string. We are experimenting with options. Users do have the alternative of using `chrome://inspect` from a desktop browser instead, though it is very slow to refresh its list and currently displays duplicates.
2. Access to media devices (eg mic) and network devices is implicitly denied by browser docks due to running on the Alloy style. Our options are to either implicitly allow, or implement a permission UX of some kind. This will likely be via a separate PR.
    * https://github.com/WizardCM/obs-browser/commit/4c65eaaf4044240f16e15a0258db3023b79945d7

Additional area I'm looking into, which may be separate PR
1. Dev Tools windows (right click Inspect from a browser dock) currently display a Chromium icon and a profile badge on Windows. To fix, this may require a change to `obs.rc.in` in the main repo if I don't successfully guess the icon ID number.

browser-app.cpp
```diff
@@ -82,16 +133,35 @@ void BrowserApp::OnBeforeCommandLineProcessing(const CefString &, CefRefPtr<CefC
 		disableFeatures += ",EnableWindowsGamingInputDataFetcher";
 #endif
 		disableFeatures += ",WebBluetooth";
+		disableFeatures += ",MediaRouter";
+		disableFeatures += ",CalculateNativeWinOcclusion";
+		disableFeatures += ",LiveCaption";
+		// https://github.com/chromiumembedded/cef/issues/3966
+		disableFeatures += ".StorageNotificationService";
```
tytan652: ?
```suggestion
		disableFeatures += ",StorageNotificationService";
```

tippfehlr:
@tytan652 I saw your :-1:, is that only related to generally using obs-browser with cef 148, or is it related to applying the patch? If it’s the latter, could you elaborate?

As far as I know, obs-browser worked before this patch, and it still works with this patch.

mihawk90:
> 2\. Access to media devices (eg mic) and network devices is implicitly denied by browser docks due to running on the Alloy style. Our options are to either implicitly allow, or implement a permission UX of some kind. This will likely be via a separate PR.

Is this not what the `--enable-media-stream` currently does? Is that removed in newer Chromium versions?

I've been using this for a while for browser docks to have access to output device listing, but AFAIK this also allows input device listing.

browser-app.cpp
WizardCM: Of course I did something this silly. Fixed.

WizardCM:
> > 2. Access to media devices (eg mic) and network devices is implicitly denied by browser docks due to running on the Alloy style. Our options are to either implicitly allow, or implement a permission UX of some kind. This will likely be via a separate PR.
> 
> Is this not what the `--enable-media-stream` currently does? Is that removed in newer Chromium versions?
> 
> I've been using this for a while for browser docks to have access to output device listing, but AFAIK this also allows input device listing.

I'm unsure if it was removed, or deprecated in favour of other things. At the very least, CEF explicitly denies the request in the Chrome Runtime's Alloy Style when it didn't in the Alloy Runtime.

We could, as a temporary solution, read the flag and manually allow those calls for now, but ultimately I think having some kind of UI for it would be better, especially as there's a proper API for it.

kurehajime:
Thanks a lot for working on this — I’m really looking forward to using a modern Chromium build in browser sources.

I have one request.
If possible, I’d like you to consider including IndexedDB in the scope of this migration.

https://github.com/WizardCM/obs-browser/blob/6d1031e0e88b31a19a0fbc96ed101cb1b1a596b3/obs-browser-plugin.cpp#L749-L755

Thanks again!

WizardCM:
> I have one request. If possible, I’d like you to consider including IndexedDB in the scope of this migration.
> 
> Thanks again!

Thanks for the feedback! I see my mistake was using a barebones scene collection with basic browser sources & docks as a testing ground - there do seem to be other directories, like IndexedDB (and probably Service Worker or WebStorage) that should be migrated too. I'll have to do some additional research on this and I'll update the PR.

panel/browser-panel.cpp
```diff
@@ -562,7 +576,16 @@ QCefCookieManager *QCefInternal::create_cookie_manager(const std::string &storag
...
 {
-	BPtr<char> rpath = obs_module_config_path(storage_path.c_str());
+	std::string path = storage_path;
+#if CHROME_VERSION_BUILD > 6533
+	// TODO: Update obs-studio to use the new path structure due to subdirectories not being supported by CEF
+	std::string legacy = "obs_profile_cookies/";
+	size_t pos = storage_path.find(legacy);
+	if (pos != std::string::npos) {
+		path.replace(pos, legacy.length(), "obs_profile_cookies_");
+	}
```
PatTheMav: See above, this should also use the helper function.

panel/browser-panel.cpp
```diff
@@ -552,8 +556,18 @@ QCefWidget *QCefInternal::create_widget(QWidget *parent, const std::string &url,
...
+	std::string path = storage_path;
+#if CHROME_VERSION_BUILD > 6533
+	// TODO: Update obs-studio to use the new path structure due to subdirectories not being supported by CEF
+	// https://github.com/obsproject/obs-studio/issues/13498
+	std::string legacy = "obs_profile_cookies/";
+	size_t pos = storage_path.find(legacy);
+	if (pos != std::string::npos) {
+		path.replace(pos, legacy.length(), "obs_profile_cookies_");
+	}
+#endif
```
PatTheMav, line 579:
```suggestion
#if CHROME_VERSION_BUILD > 6533
	// TODO: Update obs-studio to use the new path structure due to subdirectories not being supported by CEF
	// https://github.com/obsproject/obs-studio/issues/13498
	std::string path = getUpdatedCookiePath(storage_path);
#endif
```

As this code is effectively implemented twice, it should be factored out:
```C++
namespace {
std::string getUpdatedCookiePath(std::string storagePath)
{
	constexpr std::string_view searchToken{"obs_profile_cookies/};
	constexpr std::string_view replaceToken{"obs_profile_cookies_}"

	size_t pos = storage_path.find(searchToken);
	if (pos != std::string::npos) {
        storage_path.replace(pos, legacy.length(), replaceToken);
	}

	return storagePath;
}
}
```    
Which enables the suggestion at the top. The helper function gets its own copy of the referenced string by default, and I would also expect this code to be optimised via RVO, so there's only a single copy involved (the string is created in the calling function).

It's a bit unfortunate that this copy is done unconditionally, even though over time there should be less and less users for which this "upgrade" (and thus copy) is necessary, but it's a drop in the bucket for all the string copies and mutations we do in the program already.

browser-app.cpp
```diff
@@ -82,16 +133,35 @@ void BrowserApp::OnBeforeCommandLineProcessing(const CefString &, CefRefPtr<CefC
 		disableFeatures += ",EnableWindowsGamingInputDataFetcher";
 #endif
 		disableFeatures += ",WebBluetooth";
+		disableFeatures += ",MediaRouter";
+		disableFeatures += ",CalculateNativeWinOcclusion";
+		disableFeatures += ",LiveCaption";
+		// https://github.com/chromiumembedded/cef/issues/3966
+		disableFeatures += ",StorageNotificationService";
```
PatTheMav: At this point, wouldn't it be more efficient to do this:
```C++
#ifdef _WIN32
constexpr bool kIsPlatformWindows = true;
#else
constexpr bool kIsPlatformWindows = false;
#endif

[...]

void BrowserApp::OnBeforeCommandLineProcessing(const CefString &, CefRefPtr<CefCommandLine> command_line)
{

[...]

// See https://github.com/chromiumembedded/cef/issues/3966 for 'StorageNotificationService' requirement
constexpr std::string_view kDefaultDisabledFeatures{
    "HardwareMediaKeyHandling,WebBluetooth,MediaRouter,"
    "CalculateNativeWinOcclusion,LiveCaption,StorageNotificationService,"
    "EnableWindowsGamingInputDataFetcher"};

std::string disableFeatures{};

if constexpr (kIsPlatformWindows) {
    disableFeatures.append(kDefaultDisabledFeatures);
} else {
    constexpr std::string_view kWindowsFeature{",EnableWindowsGamingInputDataFetcher"};
    constexpr size_t kNonWindowsLength = kDefaultDisabledFeatures.size() - kWindowsFeature.size();
    disableFeatures.append(kDefaultDisabledFeatures.substr(0, kNonWindowsLength);
}

if (command_line->HasSwitch("disable-features")) {
    disableFeatures.append(",");
    disableFeatures.append(command_line->GetSwitchValue("disable_features"));
}

command_line->AppendSwitchWithValue("disable-features", disableFeatures);

[...]
```
The `constexpr` values are all evaluated at compile time, so the resulting code will only contain a single `disableFeatures.append(char const*, unsigned long)` call, with the correct length value computed by the compiler.

This brings the code down to a single `append` operation by default, and at most 3 `append` operations in the case of an existing `disable-features` command-line argument.

Because Windows always requires this feature to be disabled, appending it at runtime would be a permanent runtime cost that this variant avoids.

(As an added bonus, converting the preprocessor statement into `constexpr bool` and using `if constexpr` makes navigating multi-platform code easier because IDEs and language servers will now "see" the entire code again.)

WizardCM: Yeah when I was preparing this PR the same consideration had crossed my mind, but I felt it was out of scope and that I'd do it first thing after this PR is merged to make this one more straightforward to review. Let me know which you'd prefer - a separate PR post-merge or a separate commit within this PR and I'll get it done.

PatTheMav: I personally prefer the "if I'm already here making changes, I might as well clean this up and make it nice, tidy, and modern" approach (because then it's done with and not added to a pile of stuff that is usually forgotten), but @RytoEX is more selective in this regard.

I defer to him on this one.

RytoEX: Feel free to add it as a separate commit and I'll evaluate if it should be squashed or merged as a separate PR.

panel/browser-panel.cpp
WizardCM: This code path literally won't be around that long, it'll only exist for as long as it takes to merge this PR, create a PR in obs-studio, and then merge that PR - ideally never actually released. Or the check could be reversed I guess? I have made the change, let me know if it needs any additional adjustment.

browser-app.cpp
WizardCM: Added as a separate commit.

out:
Tested this on macOS arm64 (Apple silicon) against CEF 147 — browser sources render, Chrome Runtime initialises, and the Alloy→Chrome Runtime profile migration completes cleanly against a real long-lived profile. Thanks for grinding through this one.

Two macOS-specific notes:

**#517 isn't quite enough on macOS.** It bumps the plugin and the Linux helper, but `cmake/os-macos.cmake` builds four more helper executables that also compile the CEF headers, and they still fail on `std::same_as` / `derived_from` / `convertible_to`. One extra `target_compile_features(... cxx_std_20)` in the helper loop fixes it — details in https://github.com/obsproject/obs-browser/pull/517.

**The migration writes to the real profile even in portable mode.** `obs_module_config_path()` doesn't honour `--portable`, so the migration runs against `~/Library/Application Support/obs-studio/plugin_config/obs-browser` regardless. It's copy-only so nothing is destroyed, and I confirmed a stock OBS 32.1.2 (Alloy, CEF 127.0.6533.120) still loads and renders browser sources against the migrated profile afterwards — so the two can coexist on one profile. Worth mentioning to testers so they back up first, since `--portable` looks like it would isolate things but doesn't.

### Environment

| | |
|---|---|
| OS | macOS 26.5.2, arm64 |
| Toolchain | Xcode 26.6 (17F113), Apple clang 21.0.0, CMake 4.1.2 |
| CEF | `147.0.14+g76d2442+chromium-147.0.7727.138` (macosarm64), wrapper built with `USE_SANDBOX=OFF` |
| obs-studio | `0218aa658` |
| PR head | `738d9d0` |

WizardCM:
Warchamp performed some quick testing and noticed that CEF 151 (7922) crashes when interacting with browser docks. I have verified that is the case, and that CEF 150 (7871) behaves normally. I have updated the title and description to reflect this, and will separately look into why CEF 151 is misbehaving.

kurehajime:
Thanks for the huge effort on the Chrome runtime migration!

I wanted to flag a potential compatibility issue related to the permission handling that was deferred to a follow-up PR.

Chromium 142 introduced permission-gated Local Network Access (LNA) checks for requests from public HTTPS origins to local and loopback destinations ([Chrome blog](https://developer.chrome.com/blog/local-network-access)). Chromium 147 extends these checks to WebSocket connections, including loopback destinations such as `ws://127.0.0.1` ([Intent to Ship](https://groups.google.com/a/chromium.org/g/blink-dev/c/O6GMKt44Ups)).

The CEF range supported by this PR includes Chromium 147+. Since OBS browser sources and docks use Alloy style, where permission prompts are not shown by default, I'm concerned that remotely hosted HTTPS overlays connecting to local WebSocket servers may no longer be able to connect.

Connecting from a remotely hosted page to a WebSocket server running on the local machine is a fairly common pattern among streaming tools and overlays, so the impact could be fairly broad in the streaming ecosystem.

Would this be worth taking into account in the follow-up permission work, or tracking as a separate issue?

I mainly wanted to make sure it was on the radar before a CEF 147+ build ships.

WizardCM:
Thanks for the feedback @kurehajime - it'll definitely be taken into account with an initial permission implementation alongside the release of this PR. Keep an eye out for a follow-up PR, likely once this one is merged.

cmake/os-linux.cmake
```diff
@@ -14,7 +14,8 @@ add_executable(OBS::browser-helper ALIAS browser-helper)
 
 target_sources(
   browser-helper PRIVATE # cmake-format: sortable
-                         browser-app.cpp browser-app.hpp cef-headers.hpp obs-browser-page/obs-browser-page-main.cpp)
+                         browser-app.cpp browser-app.hpp browser-dummy-client.hpp browser-dummy-client.cpp
```
RytoEX:
```suggestion
                         browser-app.cpp browser-app.hpp browser-dummy-client.cpp browser-dummy-client.hpp
```

cmake/os-macos.cmake
```diff
@@ -31,7 +31,8 @@ foreach(helper IN LISTS helper_suffixes)
 
   target_sources(
     ${target_name} PRIVATE # cmake-format: sortable
-                           browser-app.cpp browser-app.hpp cef-headers.hpp obs-browser-page/obs-browser-page-main.cpp)
+                           browser-app.cpp browser-app.hpp browser-dummy-client.hpp browser-dummy-client.cpp
```
RytoEX:
```suggestion
                           browser-app.cpp browser-app.hpp browser-dummy-client.cpp browser-dummy-client.hpp
```

cmake/os-windows.cmake
```diff
@@ -13,8 +13,8 @@ add_executable(OBS::browser-helper ALIAS obs-browser-helper)
 target_sources(
   obs-browser-helper
   PRIVATE # cmake-format: sortable
-          browser-app.cpp browser-app.hpp cef-headers.hpp obs-browser-page.manifest
-          obs-browser-page/obs-browser-page-main.cpp)
+          browser-app.cpp browser-app.hpp browser-dummy-client.hpp browser-dummy-client.cpp
```
RytoEX:
```suggestion
          browser-app.cpp browser-app.hpp browser-dummy-client.cpp browser-dummy-client.hpp
```

browser-app.cpp
```diff
@@ -48,11 +54,62 @@ CefRefPtr<CefBrowserProcessHandler> BrowserApp::GetBrowserProcessHandler()
...
+
 void BrowserApp::OnRegisterCustomSchemes(CefRawPtr<CefSchemeRegistrar> registrar)
 {
 	registrar->AddCustomScheme("http", CEF_SCHEME_OPTION_STANDARD | CEF_SCHEME_OPTION_CORS_ENABLED);
 }
 
+void BrowserApp::OnContextInitialized()
+{
+	// Without a default client, CefBrowser is unmanaged, allowing full-blown Chromium windows outside of our control
+	// We don't actually want those, so define a dummy client which will automatically close any such windows
```
RytoEX:
```suggestion
	// Without a default client, CefBrowser is unmanaged, allowing full-blown Chromium windows outside of our control.
	// We don't actually want those, so define a dummy client which will automatically close any such windows.
```

browser-app.cpp
```diff
@@ -74,24 +131,39 @@ void BrowserApp::OnBeforeCommandLineProcessing(const CefString &, CefRefPtr<CefC
 		}
 	}
 
+	// See https://github.com/chromiumembedded/cef/issues/3966 for 'StorageNotificationService' requirement
```
RytoEX:
```suggestion
	// See https://github.com/chromiumembedded/cef/issues/3966 for 'StorageNotificationService' requirement.
```

browser-client.cpp
```diff
@@ -626,6 +626,17 @@ void BrowserClient::OnLoadEnd(CefRefPtr<CefBrowser>, CefRefPtr<CefFrame> frame,
 	}
 }
 
+void BrowserClient::OnLoadError(CefRefPtr<CefBrowser>, [[maybe_unused]] CefRefPtr<CefFrame> frame,
+				CefLoadHandler::ErrorCode, const CefString &, const CefString &)
+{
+#if CHROME_VERSION_BUILD > 6533
+	// CEF doesn't currently provide a way to properly disable/override the default Chrome error page
```
RytoEX:
```suggestion
	// CEF doesn't currently provide a way to properly disable/override the default Chrome error page.
```

browser-dummy-client.cpp
```diff
@@ -0,0 +1,71 @@
+
```
RytoEX, line 17: Add an appropriate license header comment block (GPLv2+).

browser-dummy-client.hpp
```diff
@@ -0,0 +1,47 @@
+#pragma once
```
RytoEX, line 18: Add an appropriate license header comment block (GPLv2+).

obs-browser-plugin.cpp
```diff
@@ -714,8 +720,122 @@ static void check_hwaccel_support(void)
...
+
+	BPtr<char> localPrefsConfigPath = obs_module_config_path("LocalPrefs.json");
+	fs::path localPrefs{localPrefsConfigPath.Get()};
+	// User has never launched old OBS Browser before
+	if (!fs::exists(localPrefs)) {
+		blog(LOG_DEBUG, "[obs-browser]: Alloy Runtime preferences not found. No migration necessary.");
+		return;
+	}
+
+	blog(LOG_INFO, "[obs-browser]: Migrating files to Chrome Runtime..");
```
RytoEX:
```suggestion
	blog(LOG_INFO, "[obs-browser]: Migrating files to Chrome Runtime...");
```

RytoEX (changes requested):
Mostly nits.

browser-dummy-client.cpp
WizardCM: Fixed

browser-dummy-client.hpp
WizardCM: Fixed

browser-app.cpp
```diff
@@ -74,24 +131,39 @@ void BrowserApp::OnBeforeCommandLineProcessing(const CefString &, CefRefPtr<CefC
...
 		// Don't override existing, as this can break OSR
-		std::string disableFeatures = command_line->GetSwitchValue("disable-features");
-		disableFeatures += ",HardwareMediaKeyHandling";
-#ifdef _WIN32
-		disableFeatures += ",EnableWindowsGamingInputDataFetcher";
-#endif
-		disableFeatures += ",WebBluetooth";
-		command_line->AppendSwitchWithValue("disable-features", disableFeatures);
+		disableFeatures.append(",");
+		disableFeatures.append(command_line->GetSwitchValue("disable_features"));
```
PatTheMav: In my testing the condition returns `true`, but the value is empty, so a single `,` is added to `disableFeatures`.

browser-app.cpp
```diff
@@ -74,24 +131,39 @@ void BrowserApp::OnBeforeCommandLineProcessing(const CefString &, CefRefPtr<CefC
...
-		command_line->AppendSwitchWithValue("disable-features", "WebBluetooth,"
-#ifdef _WIN32
-									"EnableWindowsGamingInputDataFetcher,"
-#endif
-									"HardwareMediaKeyHandling");
+		command_line->AppendSwitchWithValue("disable-blink-features", "DocumentPictureInPictureAPI");
 	}
 
 	command_line->AppendSwitchWithValue("autoplay-policy", "no-user-gesture-required");
+	command_line->AppendSwitch("disable-extensions");
```
PatTheMav, line 165: I don't see `disableFeatures` actually being used here to _actually_ disable the features, or am I missing something?

RytoEX: It looks like this line was removed and not re-added:
```cpp
command_line->AppendSwitchWithValue("disable-features", disableFeatures);
```

WizardCM: Oops, must've disappeared in the rebase or just as part of the refactor commit. Fixed.

browser-app.cpp
WizardCM: Hmm, interesting. I could extend the check to first check the length of `GetSwitchValue`? Very weird that it would do that at all, though.

RytoEX: The if statement is checking the value of:
```cpp
command_line->HasSwitch("disable-features")
```
but the append operation in the `if` statement block is using:
```cpp
command_line->GetSwitchValue("disable_features")
```

Note the swapped hyphen (`-`) and underscore (`_`). Maybe this is why the value is empty?

WizardCM: Good catch, this should be a hyphen. I've now fixed it, and Windows correctly includes the additional `AutofillActorMode`.

browser-app.cpp
```diff
@@ -74,24 +131,40 @@ void BrowserApp::OnBeforeCommandLineProcessing(const CefString &, CefRefPtr<CefC
 		}
 	}
 
+	// See https://github.com/chromiumembedded/cef/issues/3966 for 'StorageNotificationService' requirement.
+	constexpr std::string_view kDefaultDisabledFeatures{
+		"HardwareMediaKeyHandling,WebBluetooth,MediaRouter,"
+		"CalculateNativeWinOcclusion,LiveCaption,StorageNotificationService,"
```
RytoEX: Ordering nit. The last item is left "out of order" to accommodate the substring usage below.
```suggestion
		"CalculateNativeWinOcclusion,HardwareMediaKeyHandling,LiveCaption,"
		"MediaRouter,StorageNotificationService,WebBluetooth"
```

browser-app.cpp
```diff
@@ -74,24 +131,40 @@ void BrowserApp::OnBeforeCommandLineProcessing(const CefString &, CefRefPtr<CefC
...
-#ifdef _WIN32
-									"EnableWindowsGamingInputDataFetcher,"
-#endif
-									"HardwareMediaKeyHandling");
+		command_line->AppendSwitchWithValue("disable-blink-features", "DocumentPictureInPictureAPI");
 	}
 
+	command_line->AppendSwitchWithValue("disable-features", disableFeatures);
 	command_line->AppendSwitchWithValue("autoplay-policy", "no-user-gesture-required");
+	command_line->AppendSwitch("disable-extensions");
```
RytoEX: Nit, move `disable-features` below `disable-extensions`:
```suggestion
	command_line->AppendSwitchWithValue("autoplay-policy", "no-user-gesture-required");
	command_line->AppendSwitch("disable-extensions");
	command_line->AppendSwitchWithValue("disable-features", disableFeatures);
```

RytoEX (changes requested):
Nits.

tytan652:
Adding `GlobalShortcutsPortal` in disabled features for the side Linux is needed (CEF should not meddle with this portal).

browser-app.cpp
WizardCM: Done

browser-app.cpp
WizardCM: Done

PatTheMav (approved):
Checked with CEF 150 on macOS, profiles were migrated correctly as far as I could tell and new profiles (and profile data) is not visible to current master either.

WizardCM:
Added a commit based on @tytan652's request. I have verified with Chrome verbose logging that CEF doesn't complain about unrecognised entries on other platforms. A future change will be to do platform-specific implementations so we're not disabling unrecognised features on all platforms and don't overcomplicate the conditional checks in this area of the code.

RytoEX (approved):

tippfehlr:
Great to see this merged, thanks to everyone involved!

PR #528, opened 2026-06-04
Add Property Button to Inspect Sources

WizardCM:
### Description

Add an "Inspect" button to the Browser Source's Properties window to open Chrome Dev Tools.

image

This ensures feature parity between Browser Sources and Browser Docks (they have the same accessible via the context menu).

### Motivation and Context

Extensive research, testing, and rebuilding for #523 by pkv, Lina and myself has determined that Chromium have deprecated the old list page accessible from the Remote Debugging address, and restoring the functionality in any way will be a major maintenance burden going forward. This is due to the HTML file [they ship](https://github.com/chromium/chromium/blob/0cacfd015bc849186d70b11856f72e1b175c432b/content/shell/resources/shell_devtools_discovery_page.html#L40) containing a favicon that spans the entire link, assuming the file is shipped at all - CEF purposefully doesn't as part of Chrome Runtime.

Additionally, testing by myself and Warchamp confirms unreliable loading of `chrome://inspect` - on my machines it can take as long as 3 minutes to be able to even list pages, which is unacceptable.

Therefore, after thinking about how best to serve overlay developers in a meaningful way, rather than some kind of janky workaround using the Remote Debugging `/json/list` endpoint, ultimately I decided the best solution is the simplest one.

This also means users no longer have to launch OBS with `--remote-debugging-port=1234 --remote-allow-origins=http://localhost:1234` to be able to access Dev Tools, improving both security and usability.

### How Has This Been Tested?

On Windows 10, open the Properties window for a Browser Source, scroll down, and click Inspect.

### Types of changes

### Checklist:

### DIFF
--- a/obs-browser-source.cpp
+++ b/obs-browser-source.cpp
@@ -427,6 +427,19 @@ void BrowserSource::Refresh()
 	ExecuteOnBrowser([](CefRefPtr<CefBrowser> cefBrowser) { cefBrowser->ReloadIgnoreCache(); }, true);
 }
 
+void BrowserSource::Inspect()
+{
+	ExecuteOnBrowser(
+		[](CefRefPtr<CefBrowser> cefBrowser) {
+			CefWindowInfo windowInfo;
+			windowInfo.bounds.width = 900;
+			windowInfo.bounds.height = 700;
+			CefRefPtr<CefBrowserHost> host = cefBrowser->GetHost();
+			host->ShowDevTools(windowInfo, host->GetClient(), CefBrowserSettings(), {});
+		},
+		true);
+}
+
 void BrowserSource::SetBrowser(CefRefPtr<CefBrowser> b)
 {
 	std::lock_guard<std::recursive_mutex> auto_lock(lockBrowser);
--- a/obs-browser-plugin.cpp
+++ b/obs-browser-plugin.cpp
@@ -224,6 +224,15 @@ static obs_properties_t *browser_source_get_properties(void *data)
 			return false;
 		},
 		bs);
+#if CHROME_VERSION_BUILD > 6533 || defined(__APPLE__) || defined(_WIN32)
+	obs_properties_add_button2(
+		props, "inspect", obs_module_text("Inspect"),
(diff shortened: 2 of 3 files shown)


tytan652:
For Linux, guarding this behind a more recent version of CEF is required.

Only Wayland testing since CEF is already good with X11.

### With the PR as is
Non-flatpak (bugged, crash at some point):
Screenshot From 2026-06-04 16-13-55

Flatpak (no decoration):
Screenshot From 2026-06-04 16-14-55

CEF 127 is still too dependent on X11 and it really goes wrong.
Flatpak doesn't crash since it does not find an X11 server (which also explain the lack of decoration).


### With CEF 148 and also #523 #517

Non-flatpak:
Screenshot From 2026-06-04 16-26-49

Flatpak:
Screenshot From 2026-06-04 16-23-14

CEF 148 (with the chrome runtime) seems to be fine

WizardCM:
Huh, that's weird. I'll version gate it.

PatTheMav (commented):
During my testing it seems like the inspect window is another avenue for OBS to potentially crash on shutdown due to race conditions between CEF and OBS itself (crash originates in `BrowserShutdown()`).

It's probably not an issue with this PR per-se, but it means we're adding _another_ avenue for people to potentially end up with unclean shutdowns.

obs-browser-source.cpp
```diff
@@ -427,6 +427,19 @@ void BrowserSource::Refresh()
 	ExecuteOnBrowser([](CefRefPtr<CefBrowser> cefBrowser) { cefBrowser->ReloadIgnoreCache(); }, true);
 }
 
+void BrowserSource::Inspect()
+{
+	ExecuteOnBrowser(
+		[](CefRefPtr<CefBrowser> cefBrowser) {
+			CefWindowInfo windowInfo;
+			windowInfo.bounds.width = 900;
+			windowInfo.bounds.height = 700;
```
PatTheMav, line 436: Is there any logic behind those values or were those just randomly chosen?

PR #529, opened 2026-07-02
Set disable-gpu-process-crash-limit flag for CEF

notr1ch:
### Description
One of our top OBS crashers is CEF crashing the entire process if its GPU process crashes too much. Until we can figure out why the GPU is crashing so much for some users, avoid crashing OBS entirely and just have broken browser sources instead.

### Motivation and Context
Avoid CEF bringing down OBS on GPU errors.

### How Has This Been Tested?
Hasn't - I am unable to reproduce the GPU issues.

### Types of changes

### Checklist:

- [ ] My code has been tested.

### DIFF
--- a/browser-app.cpp
+++ b/browser-app.cpp
@@ -97,6 +97,9 @@ void BrowserApp::OnBeforeCommandLineProcessing(const CefString &, CefRefPtr<CefC
 #elif !defined(_WIN32)
 	command_line->AppendSwitchWithValue("ozone-platform", wayland ? "wayland" : "x11");
 #endif
+
+	// If the GPU crashes too much, don't crash OBS as well - better to just have broken sources
+	command_line->AppendSwitch("disable-gpu-process-crash-limit");
 }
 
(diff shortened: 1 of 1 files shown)


notr1ch:
Note that this setting disables GPU fallback modes too, so if HW acceleration repeatedly crashes, CEF won't fall back to software. There's no setting to disable the crashing while still enabling fallback unfortunately.

Fenrirthviti (approved):
This is probably to be a breaking change in ways we don't expect, but I'm in favor of giving it a try. At the very least, this is likely better than taking down all of OBS with it.

CEF will already be a significant testing focus, so hopefully we sort out the major things there.

Warchamp7:
As discussed off thread I still think this is an incredibly dangerous approach. The crash in question means the GPU process has failed a certain number of times and it could not fall back to software rendering.

Setting this flag will just mean Chromium can continue to try restarting the GPU process in an infinite loop which may never succeed and simply spin it's wheels forever.

Fenrirthviti:
For posterity, and per off-thread discussions I had a misunderstanding on how this actually worked in practice, and I am changing my opinion to be against it for the time being without further testing.

PR #530, opened 2026-07-28
Add per-source camera and microphone permission

Arkkis:
### Description

Added checkbox to browser source to allow browser to access camera and microphone.
image

### Motivation and Context

This allows usage of cameras and microphones via browser source easier than using launch flag: --enable-media-stream

### How Has This Been Tested?

This has been tested by building OBS Studio locally, and used webcam embedded web page for testing.

OS: CachyOS x86_64
Kernel: Linux 7.1.4-1-cachyos
Camera: Elgato Facecam MK2
OBS Studio Source version: 32.2.1

### Types of changes

### Checklist:

### DIFF
--- a/browser-client.cpp
+++ b/browser-client.cpp
@@ -71,6 +71,11 @@ CefRefPtr<CefAudioHandler> BrowserClient::GetAudioHandler()
 	return reroute_audio ? this : nullptr;
 }
 
+CefRefPtr<CefPermissionHandler> BrowserClient::GetPermissionHandler()
+{
+	return this;
+}
+
 CefRefPtr<CefRequestHandler> BrowserClient::GetRequestHandler()
 {
 	return this;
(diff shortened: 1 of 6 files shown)


WizardCM:
@Warchamp7 Would you prefer this as-is, or a dropdown similar to Page permissions, allowing the user to choose from something like

> None
> Audio only (microphone, etc)
> Video only (display, webcam, window, etc)
> Audio and video

https://github.com/WizardCM/obs-browser/commit/83b99294956854d5b59471fbae4174fbffad6f56

Or some other implementation.

Warchamp7:
I'd like to know more about the use cases this is for first.

Arkkis:
> I'd like to know more about the use cases this is for first.

My own usecase is to use webcam in web app running inside browser source.
It's not possible currently without global flag so using this kind of web app is harder for user to take into use.

Warchamp7:
> > I'd like to know more about the use cases this is for first.
> 
> My own usecase is to use webcam in web app running inside browser source. It's not possible currently without global flag so using this kind of web app is harder for user to take into use.

What sort of web app? Basically why is this needed in OBS versus simply capturing an actual proper browser instead?

Arkkis:
> > > I'd like to know more about the use cases this is for first.
> > 
> > 
> > My own usecase is to use webcam in web app running inside browser source. It's not possible currently without global flag so using this kind of web app is harder for user to take into use.
> 
> What sort of web app? Basically why is this needed in OBS versus simply capturing an actual proper browser instead?

Since it's webcam overlay, I think browser source is better fit than capturing browser when you can just drop link into OBS and it works right away. This is simply the easiest experience for user.

Warchamp7:
> > > > I'd like to know more about the use cases this is for first.
> > > 
> > > 
> > > My own usecase is to use webcam in web app running inside browser source. It's not possible currently without global flag so using this kind of web app is harder for user to take into use.
> > 
> > 
> > What sort of web app? Basically why is this needed in OBS versus simply capturing an actual proper browser instead?
> 
> Since it's webcam overlay, I think browser source is better fit than capturing browser when you can just drop link into OBS and it works right away. This is simply the easiest experience for user.

What sort of webcam overlay? Can you please provide an actual example?

Arkkis:
> > > > > I'd like to know more about the use cases this is for first.
> > > > 
> > > > 
> > > > My own usecase is to use webcam in web app running inside browser source. It's not possible currently without global flag so using this kind of web app is harder for user to take into use.
> > > 
> > > 
> > > What sort of web app? Basically why is this needed in OBS versus simply capturing an actual proper browser instead?
> > 
> > 
> > Since it's webcam overlay, I think browser source is better fit than capturing browser when you can just drop link into OBS and it works right away. This is simply the easiest experience for user.
> 
> What sort of webcam overlay? Can you please provide an actual example?

There you go (WIP): https://efekti.arkkis.com/studio

Warchamp7:
> There you go (WIP): https://efekti.arkkis.com/studio

I'm still not seeing a reasonable justification for why this _needs_ to be implemented as a browser source instead of a separate page or even standalone app.

Issue #488, opened 2025-06-13
Frame swap artifacts occurring when HW acceleration enabled and CPU under load


Dan-Shields:
### Operating System Info

Windows 11

### Other OS

Windows 11 & Windows 10

### OBS Studio Version

31.1.0-beta2

### OBS Studio Version (Other)

31.1.0-beta2 & 31.0.3

### OBS Studio Log URL

https://obsproject.com/logs/HsUwuFk33D2Y1Yxw

### Expected Behavior

The browser source should be rendered artifact-free.

### Current Behavior

Below are screencaps showing the 1-frame long artifacts affecting a browser source set to play a 4k60 video above a #00ff00 color source. There are patterns across the different screencaps, but the artifacts are not always the same. These images were captured by screen capturing the OBS preview, but can also be seen on recordings made by the OBS doing the rendering, as well as in decklink output.

Browser hardware acceleration must be enabled for the issue to arise.

The frequency of frames which are effected is proportional to the load that the CPU is under. The CPU core assigned to run OBS compositing is particularly sensitive.

The issue appears either infrequently or not-at-all if the GPU is running 100% (i.e. GPU bottlenecked). In this instance the video will slow/drop frames, but no artificing is minimized. This part was tested changing videos to ones that were hardware decoded vs software, higher and lower bitrates, and different resolutions to affect GPU usage.

The above was found by correlating HW usage in task manager while trying various combos of things we found to make the problem worse.

Things that exacerbate the issue:
- setting the OBS process priority to idle and unfocusing the window
- throttling the CPU
- applying load across the CPU (tested running a multi-core CPU [CineBench 2024](https://www.maxon.net/en/downloads/cinebench-2024-downloads) benchmark)
- starting a recording using the software encoder (x264)
- starting obs virtual camera output
- starting decklink output

Things that don't seem to affect the issue:
- pixel format (happens on BGRA & NV12 at least)
- drop frame (happens on 60 & 59.94 fps)
- starting a recording using a GPU encoder (NVENC)
- the specific contents of the browser source, as long as its updating each frame (we have found the same issue with CSS animation-based pages, for example)

The issue was first noticed to happen occasionally on beefy renderer machines with identical hardware to each other, then reproduced on several separate machines, one of which is where the attached logs come from. On those we see several affected frames per second.

We have some renderer machines on slightly different hardware, but running the same software which don't show the issue at all.

[...]

PR #531, opened 2026-08-04
Fix browser source tearing with hardware acceleration

hubert-blasttv:
### Description

Since the CEF M124 OSR rework, `OnAcceleratedPaint` hands out a texture from a small internal pool, with no keyed mutex. The handle ["cannot be cached and cannot be accessed outside of this callback"](https://github.com/chromiumembedded/cef/blob/af8ada6a3ab9878669a3a6c46d9e0402279ac068/include/cef_render_handler.h#L163-L166) and the contents "should be copied to a texture owned by the client application". ["Resources will be released to the underlying pool for reuse when the callback returns"](https://github.com/chromiumembedded/cef/blob/af8ada6a3ab9878669a3a6c46d9e0402279ac068/include/internal/cef_types_win.h#L130-L131) and that the texture "is instantiated without a keyed mutex".

The current implementation does the opposite: open the handle, keep the resource as `bs->texture`, and let the render thread sample it during composition at some later point. When the compositor falls behind under CPU load, CEF is already writing a later frame into that same pooled texture while we sample it, which shows up as the one-frame swaps and tears from the issue.

#### Fix

`CopyAcceleratedTexture()` now copies the frame into a texture we own, inside the callback, while the handle is guaranteed valid. The copy target is reused across frames and only reallocated when the size or format changes. `gs_flush()` submits the copy before the callback returns, since the pooled texture can be reused after that. The render thread then samples our private, stable copy.

The `gs_texture_acquire_sync(shared, 1, 50)` is a no-op right now but it cooperates, with a bounded wait, if a future producer exposes a keyed mutex.

One note: the flush submits the copy but cannot fence it, because CEF exposes no fence through this API. This is as close to the documented contract as a consumer can get. The exposure window shrinks from a full compositor frame of sampling to command submission granularity, and the pool depth gives the copy several frame intervals of slack in practice.

### Motivation and Context
Recurrence of #488 

### How Has This Been Tested?

Retested multiple times against example from #488, and with python script to detect tear artifacts.

### Types of changes

### Checklist:

### DIFF
--- a/browser-client.cpp
+++ b/browser-client.cpp
@@ -369,6 +369,45 @@ void BrowserClient::UpdateExtraTexture()
 	}
 }
 
+#ifdef _WIN32
+/* CEF's shared texture is pooled and only valid inside this callback;
+ * copy it into a texture we own instead of sampling it later (#488).
+ * Keyed-mutex sync was disabled "for now" in eced8c7 (2022), and since
+ * the CEF M124 rework the pooled textures have no mutex at all. */
+void BrowserClient::CopyAcceleratedTexture(void *shared_handle)
+{
+	gs_texture_t *shared = gs_texture_open_nt_shared((uint32_t)(uintptr_t)shared_handle);
+	if (!shared)
+		return;
+
+	/* No-op today (no keyed mutex); bounded wait for future producers. */
+	const bool acquired = gs_texture_acquire_sync(shared, 1, 50) == 0;
+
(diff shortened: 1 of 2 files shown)


Dan-Shields:
I have independently tested the patch and this solves our issue!

Dan-Shields:
I've confirmed that this patch builds and works as intended with the CEF 150 update ([obs-studio/#13900](https://github.com/obsproject/obs-studio/pull/13900))

